PR #3, opened 2015-02-16
default view template shouldn't fail when assignee is nil

jaybuff:
The default view template was failing with this error whenever I tried to
view a jira that had no assignee:
assignee: 2015-02-16T08:31:58.564-08:00 ERROR [util.go:109] Failed to execute template: template: template:7:20: executing "template" at <.fields.assignee.nam...>: nil pointer evaluating interface {}.name

### DIFF
--- a/jira/cli/templates.go
+++ b/jira/cli/templates.go
@@ -26,7 +26,7 @@ summary: {{ .fields.summary }}
 project: {{ .fields.project.key }}
 components: {{ range .fields.components }}{{ .name }} {{end}}
 issuetype: {{ .fields.issuetype.name }}
-assignee: {{ .fields.assignee.name }}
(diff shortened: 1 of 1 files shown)


jira/cli/templates.go
```diff
@@ -26,7 +26,7 @@ summary: {{ .fields.summary }}
 project: {{ .fields.project.key }}
 components: {{ range .fields.components }}{{ .name }} {{end}}
 issuetype: {{ .fields.issuetype.name }}
-assignee: {{ .fields.assignee.name }}
+assignee: {{ if .fields.assignee }}{{ .fields.assignee }}{{end}}
```
coryb: almost, the inner block should be {{ .fields.assignee.name }}

PR #7, opened 2015-02-18
validate project

jaybuff:
$ jira create
panic: runtime error: index out of range

$ curl -s 'https://issues.apache.org/jira/rest/api/2/issue/createmeta?projectKeys=mesos&issuetypeNames=Bug&expand=projects.issuetypes.fields' |jq .
{
  "projects": [],
  "expand": "projects"
}

I read half a go book like two years ago, so...

### DIFF
--- a/jira/cli/commands.go
+++ b/jira/cli/commands.go
@@ -220,6 +220,11 @@ func (c *Cli) CmdCreateMeta(project string, issuetype string) error {
 	}
 
 	if val, ok := data.(map[string]interface{})["projects"]; ok {
+		if len(val.([]interface{})) == 0 {
+			err = fmt.Errorf("Project '%s' or issuetype '%s' unknown.  Unable to createmeta.", project, issuetype)
+			log.Error("%s", err)
+			return err
+		}
 		if val, ok = val.([]interface{})[0].(map[string]interface{})["issuetypes"]; ok {
 			data = val.([]interface{})[0]
 		}
@@ -253,6 +258,11 @@ func (c *Cli) CmdCreate(project string, issuetype string) error {
 	issueData["overrides"].(map[string]string)["issuetype"] = issuetype
 
 	if val, ok := data.(map[string]interface{})["projects"]; ok {
+		if len(val.([]interface{})) == 0 {
+			err = fmt.Errorf("Project '%s' or issuetype '%s' unknown.  Unable to create issue.", project, issuetype)
+			log.Error("%s", err)
+			return err
+		}
 		if val, ok = val.([]interface{})[0].(map[string]interface{})["issuetypes"]; ok {
 			issueData["meta"] = val.([]interface{})[0]
 		}
--- a/jira/cli/cli.go
+++ b/jira/cli/cli.go
@@ -32,6 +32,10 @@ func New(opts map[string]string) *Cli {
 	endpoint, _ := opts["endpoint"]
 	url, _ := url.Parse(strings.TrimRight(endpoint, "/"))
 
+	if project, ok := opts["project"]; ok {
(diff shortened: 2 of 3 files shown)


coryb:
You sure?  issuetype is a required field I think on all jira issues.  My hunch is that the createmeta politely returns an empty list when you are unauthenticated?

try jira login, then jira create?

-Cory

coryb:
If the `jira login` fixes your issue, then the easiest thing to do is probably just a add a **c.CmdLogin** in main.go just before the **c.CmdCreate**

jaybuff:
Nope, still an issue:

    $ jira login
    Enter Password for jaybuff:

    $ jira create 
    panic: runtime error: index out of range

To double check, my browser is logged in to issues.apache.org (it has the cookie) and I get the same json with an empty projects back from createmeta when I do that get with my browser.

coryb:
Hmm, there is something wrong, without createmeta we cannot create/validate issues. I wonder if the expand it wrong.

what do you get with `jira createmeta -p mesos`

coryb:
Oh, duh ... the project key is case sensitive, set it to "MESOS", not "mesos"

jaybuff:
Ah-ha!  So I guess the issue is we need to validate project.

jira/cli/commands.go
```diff
@@ -253,9 +253,11 @@ func (c *Cli) CmdCreate(project string, issuetype string) error {
 	issueData["overrides"].(map[string]string)["issuetype"] = issuetype
 
 	if val, ok := data.(map[string]interface{})["projects"]; ok {
-		if val, ok = val.([]interface{})[0].(map[string]interface{})["issuetypes"]; ok {
-			issueData["meta"] = val.([]interface{})[0]
-		}
+        if len(val.([]interface{})) > 0 {
```
coryb: okay, yeah, I think we need validation here.  Something like:

```go
if len(val.([]interface{}) == 0 {
  err = fmt.Errorf("Project %s or issuetype %s unknown, unable to create issue", project, issuetype)
  log.Error(err)
  return err
}
```

jaybuff: +1 I'll update PR.

jaybuff:
We need this validation on all commands that take project:

    $ jira  |grep -i project
    2015/02/18 17:00:39 Failed to parse options:
      jira [-v ...] [-u USER] [-e URI] [-t FILE] (ls|list) ( [-q JQL] | [-p PROJECT] [-c COMPONENT] [-a ASSIGNEE] [-i ISSUETYPE] [-w WATCHER] [-r REPORTER])
      jira [-v ...] [-u USER] [-e URI] [-b] [-t FILE] create [--noedit] [-p PROJECT] [-i ISSUETYPE] [-o KEY=VAL]...
      jira [-v ...] [-u USER] [-e URI] [-t FILE] issuetypes [-p PROJECT]
      jira [-v ...] [-u USER] [-e URI] [-t FILE] createmeta [-p PROJECT] [-i ISSUETYPE]

coryb:
Meh, we dont really need validation except for create.  The list, issuetypes and createmeta commands will just return no data with invalid input.  We can force `strings.ToUpper(project)` in the code for some validation.  The one that is critical to validate is the create command.

-Cory

jaybuff:
Well, createmeta panics:

    $ jira createmeta
    panic: runtime error: index out of range

and ls gives me no meaningful error:

    $ jira ls -p adsasdf
    2015-02-18T17:42:18.985-08:00 ERROR [cli.go:158] response status: 400 Bad Request

Sure, we can ignore issuetypes.  Plus I've already written the code :)

coryb:
Cool, looks good, thanks.
-Cory

PR #9, opened 2015-07-01
fix "take" command not honouring user option

nelfin:
"take" was simply a partial on "assign", but accidentally used the value
of $USER from the environment rather than `opts["user"]`, preventing the
user from overriding this value in config.yml or as a command-line
argument.


coryb:
Nice, thanks!

It sounds like your USER environment variable does not match the your user for jira?  You can change the default "USER" in your ~/.jira.d/config.yml so you wont have to use the --user option much.
```bash
user: coryb
editor: emacs -nw
endpoint: https://jira.company.com
```

PR #11, opened 2015-07-31
Add --max_results option for 'ls'

### LINKED ISSUE #10: CmdList is hardcoded to 500 results.
Noticed this whilst looking thru the commands.go

````
json, err := jsonEncode(map[string]interface{}{¬
     "jql":        query,
     "startAt":    "0",
     "maxResults": "500",
     "fields":     fields,
})
````

I know we have projects with more than this open, so it would be great to expose this as an option.

(NB: i plan to submit a PR for this!)

mikepea:
This closes #10

Shifts the hardcoded maxResults value for the cmdList json body
into a 'max_results' option.

Note that testing against our JIRA instance, in a project with
more than 1000 open issues, suggests that the JIRA has an internal
limit of 1000 results in a single query.

### DIFF
--- a/jira/cli/commands.go
+++ b/jira/cli/commands.go
@@ -116,7 +116,7 @@ func (c *Cli) CmdList() error {
 	json, err := jsonEncode(map[string]interface{}{
 		"jql":        query,
 		"startAt":    "0",
-		"maxResults": "500",
+		"maxResults": c.opts["max_results"],
 		"fields":     fields,
 	})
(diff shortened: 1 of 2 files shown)


coryb:
Looks good, thanks!

coryb:
This has been put into a 0.0.8 release:
https://github.com/Netflix-Skunkworks/go-jira/releases/tag/0.0.8

PR #13, opened 2015-10-02
update default edit template, and copy the overrides into the opts

pepin:
The default edit template fails when trying examples in the README.  Copying the overrides into the opts map ensures that the values specified there are present in the template as .overrides.$name, instead of .overrides.overrides.$name.

I made the comment optional, and added an optional label to the default template, so you can just update a label for an issue if you want.

Another way to accomplish this change would be to just pass the opts["overrides"] to the template instead of the full map.  My guess is that was changed at some point when the -m paramater was added. after the docs were written, and everyone has their own custom template files at this point.  This seemed less potentially impactful, since the templates may be using values already from opts.

### DIFF
--- a/jira/cli/templates.go
+++ b/jira/cli/templates.go
@@ -25,7 +25,7 @@ const default_list_template = "{{ range .issues }}{{ .key | append \":\" | print
 const default_table_template = `+{{ "-" | rep 16 }}+{{ "-" | rep 57 }}+{{ "-" | rep 14 }}+{{ "-" | rep 14 }}+{{ "-" | rep 12 }}+{{ "-" | rep 14 }}+{{ "-" | rep 14 }}+
 | {{ "Issue" | printf "%-14s" }} | {{ "Summary" | printf "%-55s" }} | {{ "Priority" | printf "%-12s" }} | {{ "Status" | printf "%-12s" }} | {{ "Age" | printf "%-10s" }} | {{ "Reporter" | printf "%-12s" }} | {{ "Assignee" | printf "%-12s" }} |
 +{{ "-" | rep 16 }}+{{ "-" | rep 57 }}+{{ "-" | rep 14 }}+{{ "-" | rep 14 }}+{{ "-" | rep 12 }}+{{ "-" | rep 14 }}+{{ "-" | rep 14 }}+
-{{ range .issues }}| {{ .key | printf "%-14s"}} | {{ .fields.summary | abbrev 55 | printf "%-55s" }} | {{.fields.priority.name | printf "%-12s" }} | {{.fields.status.name | printf "%-12s" }} | {{.fields.created | age | printf "%-10s" }} | {{.fields.reporter.name | printf "%-12s"}} | {{if .fields.assignee }}{{.fields.assignee.name | printf "%-12s" }}{{else}}<unassigned>{{end}} |
+{{ range .issues }}| {{ .key | printf "%-14s"}} | {{ .fields.summary | abbrev 55 | printf "%-55s" }} | {{.fields.priority.name | printf "%-12s" }} | {{.fields.status.name | printf "%-12s" }} | {{.fields.created | age | printf "%-10s" }} | {{if .fields.reporter}}{{ .fields.reporter.name | printf "%-12s"}}{{else}}<unassigned>{{end}} | {{if .fields.assignee }}{{.fields.assignee.name | printf "%-12s" }}{{else}}<unassigned>{{end}} |
 {{ end }}+{{ "-" | rep 16 }}+{{ "-" | rep 57 }}+{{ "-" | rep 14 }}+{{ "-" | rep 14 }}+{{ "-" | rep 12 }}+{{ "-" | rep 14 }}+{{ "-" | rep 14 }}+
 `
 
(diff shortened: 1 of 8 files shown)


jira/cli/templates.go
```diff
@@ -52,12 +52,18 @@ comments:
 `
 const default_edit_template = `# issue: {{ .key }}
 update:
+{{if .overrides.comment}}
```
coryb: The goal here is to auto populate the comment if someone used -m "my comment" but still allow someone to manually type in a comment while running "jira edit".  So in the case that someone wants to add a comment but does not use "jira edit -m ..." that section of the template needs to exist, so I think it should still look something like:
```
comment:
  - add:
    body: |~
       {{ or .overrides.comment "" | indent 10 }}
```

jira/cli/templates.go
```diff
@@ -52,12 +52,18 @@ comments:
...
   comment:
-    - add: 
+    - add:
         body: |~
           {{ or .overrides.comment "" | indent 10 }}
+{{end}}
+{{ if .overrides.label }}
```
coryb: I think this is probably similar to comment, someone can use `jira edit -o label=foobar` but maybe they should see a labels section if no overrides are present?  So I would expect this to look something like:
```
labels:{{range label in .fields.labels}
  - add: {{label}}{{end}}
  - add: {{ or .overrides.label "" }}
```

(I am assuming lables pre-exist in an issue, so perhaps we show users current labels and allow them to add more?)

jira/cli/templates.go
```diff
@@ -52,12 +52,18 @@ comments:
...
+  labels:
+    - add: {{ .overrides.label }}
+{{end}}
 fields:
-  summary: {{ or .overrides.summary .fields.summary }}
+  summary: "{{ or .overrides.summary .fields.summary }}"
```
coryb: Curious on the rational here?  The quotes seem unnecessary unless you want to add meta characters to the summary?  Perhaps a summary of "foo: bar" will throw off the yaml parser?

coryb:
Thanks for the PR, I think you are right in that the overrides got lost when I did some recent refactoring.

jira/cli/templates.go
pepin: I think thats fine, the reason I changed it was because it doesn't allow you to do anything without a comment, but thats my case and I could just update my own templates locally.

When I was looking around in there, it seems that our version of jira validates the comment is not just whitespace, and gives us back an error.

jira/cli/templates.go
pepin: Yeah, thats a good Idea, What happens if you leave the -add: line empty there?

Is the program supposed to ignore it if its empty?  I noticed it does not do that for the comment section, but maybe it is supposed to?

jira/cli/templates.go
pepin: You are right.  The summary fields sometimes contain chars like ] or [  ie '[something]'  in the summary. 

This is another case where I could use a custom template, but I felt that as a total n00b coming in here, these things didn't quite work for me, and I may have just gone off frustrated to some other project that worked out of the box.  Thats why I thought the default templates should 'just work' for people who don't really understand how it works yet.  

The advanced user or someone who wants to dig into the source could very easily write up a template that works exactly as they want it to.  I had thought about adding a quoting function, whereby we could determine if there were meta characters in the string and escape them, but opted not to as this was quicker and seems to work for most cases.

Now that I think of it, it may be simple enough to just build a temporary yaml string and see if it parses as a string.  If not, then add the quotes, or make it a multi-line like we have above for comments.   So you'd have something like 

    summary: {{ or .overrides.summary .fields.summary | quote }}

And you'd get 

    summary: >-
             my [summary] that contains meta chars and possibly "quotes"

I did quick check in python and it parses there:

    >>> yaml.load("foo:\n  summary: >-\n    my [summary] that contains meta chars and possibly \"quotes\"")
    {'foo': {'summary': 'my [summary] that contains meta chars and possibly "quotes"'}}

jira/cli/templates.go
coryb: Interesting.  Users should be able to edit an issue without adding a comment (change watchers, change assigned, etc).  If you leave the comment body blank go-jira should strip that field from the document submitted to the jira api.  There might be a bug with that, the document is fixed in the yamlFixup routine in util.go.

jira/cli/templates.go
coryb: In theory the yamlFixup routine should strip out the empty `add: ` line, if it is not doing that there is a bug.

jira/cli/templates.go
coryb: Okay make sense.  I guess I prefer this syntax:
```
summary: >-
    {{ or .overrides.summary .fields.summary }}
```

I have tried to move all extraneous formatting characters (ie quotes, etc)

jira/cli/templates.go
coryb: I just found a bug that would try to post "\n" for the comment body if you don't modify the empty template field, which i think is causing this issue for you.  I have fixed it with [47ced2f].

Also if you want to see what go-jira would post without actually posting data, you can use `--dryrun --verbose --verbose` which will now print the json data we would have posted, without actually posting the document.  It can be helpful to debug stuff like this.

coryb:
Finally closing this issue/PR, template have been update to use:
```
summary: >-
   stuff [here]
```

PR #14, opened 2015-10-15
Add 'updated' to queryfields; dateFormat template command; bug fix.

mikepea:
I've included the 'updated' field into the default query fields, as it seemed like an oversight, and I need it :)

In the process, noticed that export-templates was broken due to a typo, so fixed that.

I've added a dateFormat template helper, so I can get the create/update time in a more report friendly form.

### DIFF
--- a/README.md
+++ b/README.md
@@ -211,7 +211,7 @@ General Options:
 Query Options:
   -a --assignee=USER        Username assigned the issue
   -c --component=COMPONENT  Component to Search for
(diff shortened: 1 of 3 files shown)


coryb:
Nice, looks good, thanks!

Side note, you could have added "updated" to your default list of queryfields via configuration by modifying your ~/.jira.d/config.yml to have:
```
queryfields: summary,created,updated,priority,status,reporter,assignee
```
-Cory

PR #20, opened 2015-12-23
Add a 'join' func to the template engine

mikepea:
I needed this so that I can display JIRA labels in my 'list' template, but it should be generally useful.

Happy to have this pushed back for a more robust implementation with tests.

### DIFF
--- a/util.go
+++ b/util.go
@@ -161,6 +161,13 @@ func runTemplate(templateContent string, data interface{}, out io.Writer) error 
 		"split": func(sep string, content string) []string {
 			return strings.Split(content, sep)
 		},
+		"join": func(sep string, content []interface{}) string {
+			vals := make([]string, len(content))
+			for i, v := range content {
+				vals[i] = v.(string)
+			}
+			return strings.Join(vals, sep)
+		},
 		"abbrev": func(max int, content string) string {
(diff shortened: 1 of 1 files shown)


mikepea:
Note that to enable displaying labels in 'list' output, you need to append to queryfields:

    queryfields: "summary,created,updated,priority,status,reporter,assignee,labels"

coryb:
Looks good.  An alternative to using join would be something like this:
```
labels: {{ range .fields.labels }}
  - {{ . }}{{end}}
```
and just iterate over the labels and print them out, one per line.
-Cory

coryb:
Oh, I see, you wanted this for "list", I was thinking "view", so you obviously don't want to print multiline stuff in "list".   So ignore my comment :)

Thanks,
-Cory

lvh:
Uh, how do I use this? I tried to do the following:

```yaml
  - name: search
    help: search for text
    script: echo jira list --query='text ~ \"{{ join " " args.ARGS }}\"'
    args:
      - name: ARGS
        repeat: true
```

But got: `ERROR Invalid Usage: template: gojira:1:45: executing "gojira" at <args>: wrong type for value; expected []interface {}; got []string`

PR #21, opened 2015-12-24
Add 'labels' command to set/add/remove labels

mikepea:
This adds 'labels' command, which allows for the setting, addition and removal
of labels on an issue.

'set' action resets the issue labels to the list provided.
'add' action adds the supplied labels to the issue
'remove' action removes the supplied labels from the issue

The API already gracefully handles duplication, removal of non-existant labels, and the
supplying of an empty list of labels (which is useful in the case of 'set')

Eg

jira labels TEST-123 add label1 label2 label3
jira labels TEST-123 remove label1 label2
jira labels TEST-123 set label1 label2 label3
jira labels TEST-123 set # clears any labels on the issue
jira labels TEST-123 add # no-op
jira labels TEST-123 remove # no-op

This mirrors the functionality of the API.

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -542,7 +542,7 @@ func (c *Cli) CmdComment(issue string) error {
 	return nil
 }
 
-func (c *Cli) CmdLabels(issue string, command string, labels []string) error {
+func (c *Cli) CmdLabels(command string, issue string, labels []string) error {
 	log.Debug("label called")
 
 	if command != "add" && command != "remove" && command != "set" {
--- a/commands.go
+++ b/commands.go
@@ -544,6 +544,71 @@ func (c *Cli) CmdComment(issue string) error {
 	return nil
 }
 
+func (c *Cli) CmdLabels(issue string, command string, labels []string) error {
+	log.Debug("label called")
+
+	if command != "add" && command != "remove" && command != "set" {
+		return fmt.Errorf("command must be 'add', 'set' or 'remove': %q is invalid", command)
+	}
+
+	handlePut := func(json string) error {
+		log.Debug("JSON: %s", json)
+		uri := fmt.Sprintf("%s/rest/api/2/issue/%s", c.endpoint, issue)
+		if c.getOptBool("dryrun", false) {
+			log.Debug("PUT: %s", json)
+			log.Debug("Dryrun mode, skipping POST")
+			return nil
+		}
+		resp, err := c.put(uri, json)
+		if err != nil {
+			return err
+		}
+
+		if resp.StatusCode == 204 {
+			c.Browse(issue)
+			if !c.opts["quiet"].(bool) {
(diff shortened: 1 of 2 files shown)


coryb:
Hi Mike,

I am tying to figure out how to proceed with this PR.  I appreciate that the jira api is reflected cleanly in the usage, but to me the usage is too disjoint compared to the other go-jira commands. 

how about some operational commands like:
`jira add-labels TEST-123 label1 [label2]...`
`jira rm-labels TEST-123 label1 [label2]...`
`jira set-labels TEST-123 label1 [label2]...`

Or perhaps something like:

`jira add labels TEST-123 label1 [label2]...`
`jira rm labels TEST-123 label1 [label2]...`
`jira set labels TEST-123 label1 [label2]...`

The second group is a little trickier on the implementation side, but opens us up for consistent usage on other set fields, something like:
`jira add components ...`
`jira set watchers ...`

We probably need to expose labels in the main `edit` templates which would allow you to add/remove/set labels via the edit operation.  I think if we add something like this to the edit template it would all users to rm/set via the edit template, and automatically add via `jira edit TEST-123 -o labels label1,label2` --noedit:

```
{{if .meta.fields.labels}}
  labels: {{range .fields.labels}}
    - {{.}}{{end}}{{if .overrides.labels}}{{range (split "," .overrides.labels)}}
    - {{.}}{{end}}{{end}}
    - {{end}}
```

-Cory

mikepea:
Totally appreciate that this is a bit of a curve-ball for the go-jira
interface, and am happy to rework however you feel is best.

I'll take a look at how I can work it as per your second suggestion - I
agree that this makes sense.

I've not thought too much about the edit interface at all yet, but will
take a look at how it works. I think I get what you're saying though - that
with the right template it should 'just work' - based on the fields-method
of update?

coryb:
Okay, sounds good.

Thinking about it, this usage is pretty close to what you already have:
`jira add labels TEST-123 label1 [label2]...`
`jira rm labels TEST-123 label1 [label2]...`
`jira set labels TEST-123 label1 [label2]...`

So I think that will actually be trivial to tweak, you just have to reorder the arguments passed to CmdLabel.  With this usage the `command` variable should already be set to `labels` since go-jira looks for commands in the first or second argument position (ie to allow for `jira ISSUE-1 blocks ISSUE-2`).

The edit interface is a bit of a tangent and not really relevant to this PR.  Basically `jira edit ...` will open up the jira issue in your favorite editor and let you modify yaml directly.  The snippet I put above will include a section for labels if that is a field in the available edit fields.  It will list the current labels, append any labels specified with a `-o labels=label1,label2` option, and then add a dangling `-` to allow you to just type them in in the editor.

-Cory

mikepea:
@coryb updated as per our discussion :)

coryb:
Looks good, thanks!
-Cory

PR #22, opened 2015-12-31
Expose key functionality for library consumption

mikepea:
I've added exposed versions of runTemplate and getTemplate, so I can access the template engine with my own Buffer. I opted to add tandem RunTemplate and GetTemplate, rather than a search-replace fix, to make the smallest change possible. Happy to rework if you're not fussed about changing the internal API.

I've split the API-integrated part of `CmdView` into `ViewIssue(...)` as per `FindIssues(...)`.

These are the only exposures I need right now, let me know if you'd like anything else included.

### DIFF
--- a/cli.go
+++ b/cli.go
@@ -177,6 +177,10 @@ func (c *Cli) makeRequest(req *http.Request) (resp *http.Response, err error) {
 	return resp, nil
 }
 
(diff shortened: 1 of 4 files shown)


coryb:
looks good to me, thanks!
-Cory

Issue #23, opened 2016-01-04
default edit template is problematic


mikepea:
The standard permissions set on our Jira instance do not allow for update of `reporter`. I'm fairly confident this will be a commonplace limitation for unprivileged users.

This means that every 'edit' operation initially fails with:

    {"errorMessages":[],"errors":{"reporter":"Field 'reporter' cannot be set. It is not on the appropriate screen, or unknown."}}

I appreciate that a custom edit template can be provided to remove this, but in my mind it makes more sense for the default edit template to fit a lower common denominator.

coryb:
Closing this issue, the edit and create templates were modfied to only show reporter if it is a valid metadata field.

-Cory

PR #24, opened 2016-01-04
Reporter is not generally editable.

mikepea:
Support the lowest common denominator for default_edit_template, as per #23

`reporter` is not usually editable by unprivileged users. Even if it is left as-is,
the PUT request is trying to update it, resulting in the following error:

    {"errorMessages":[],"errors":{"reporter":"Field 'reporter' cannot be set. It is not on the appropriate screen, or unknown."}}

Commenting out the field leaves the information visible, so is a reasonable compromise.

### DIFF
--- a/templates.go
+++ b/templates.go
@@ -64,8 +64,8 @@ fields:
     - name: {{ .name }}{{end}}{{end}}
   assignee:
     name: {{ if .overrides.assignee }}{{.overrides.assignee}}{{else}}{{if .fields.assignee }}{{ .fields.assignee.name }}{{end}}{{end}}
-  reporter:
-    name: {{ if .overrides.reporter }}{{ .overrides.reporter }}{{else if .fields.reporter}}{{ .fields.reporter.name }}{{end}}
+#	reporter:
+#	  name: {{ if .overrides.reporter }}{{ .overrides.reporter }}{{else if .fields.reporter}}{{ .fields.reporter.name }}{{end}}
   # watchers
   customfield_10110: {{ range .fields.customfield_10110 }}
     - name: {{ .name }}{{end}}{{if .overrides.watcher}}


coryb:
Okay, yeah, most fields probably need to be checked before showing them as editable.  I doubt "customfield_10110" is "watchers" everywhere also.  It just happens to be so on my system.  

Anyway, before we start populating the edit template we call the /editmeta REST api to get a list of editable fields.  That data is available in .meta.  So I think we just need to add a if statement around the reporter section, like:

```
{{if .meta.fields.reporter}}...{{end}}
```

So I think the overall diff would look like this:
```diff
--- a/templates.go
+++ b/templates.go
@@ -63,9 +63,9 @@ fields:
     - name: {{.}}{{end}}{{else}}{{ range .fields.components }}
     - name: {{ .name }}{{end}}{{end}}
   assignee:
-    name: {{ if .overrides.assignee }}{{.overrides.assignee}}{{else}}{{if .fields.assignee }}{{ .fields.assignee.name }}{{end}}{{end}}
+    name: {{ if .overrides.assignee }}{{.overrides.assignee}}{{else}}{{if .fields.assignee }}{{ .fields.assignee.name }}{{end}}{{end}}{{if .meta.fields.reporter}}
   reporter:
-    name: {{ if .overrides.reporter }}{{ .overrides.reporter }}{{else if .fields.reporter}}{{ .fields.reporter.name }}{{end}}
+    name: {{ if .overrides.reporter }}{{ .overrides.reporter }}{{else if .fields.reporter}}{{ .fields.reporter.name }}{{end}}{{end}}
   # watchers
   customfield_10110: {{ range .fields.customfield_10110 }}
     - name: {{ .name }}{{end}}{{if .overrides.watcher}}
```

That should not present the reporter section of the template unless it is editable by you.

The `{{if ..}}` starts on the previous line so the whitespace looks normal when printing the template.

-Cory

PR #26, opened 2016-01-13
Add 'vote' and 'unvote'

mikepea:
This adds support for voting on issues via CmdVote() and CmdUnvote()

Voting on issues is always done as the logged in user, it appears you
can't case a vote for another user:

https://docs.atlassian.com/jira/REST/latest/#api/2/issue-addVote

This required adding a cli.delete() handler, naturally with no content
(as per RFC2616)

This is ripe for DRY-ing out, but I will leave that for a future PR.

Worth noting is that you cannot vote for your own issues, this results in:

    2016-01-13T21:35:41.315Z ERROR [cli.go:184] response status: 404 Not Found
    2016-01-13T21:35:41.315Z ERROR [commands.go:439] Unexpected Response From POST:
    {snip}
    {"errorMessages":["You cannot vote for an issue you have reported."],"errors":{}}

### DIFF
--- a/cli.go
+++ b/cli.go
@@ -106,6 +106,24 @@ func (c *Cli) put(uri string, content string) (*http.Response, error) {
 	return c.makeRequestWithContent("PUT", uri, content)
 }
 
+func (c *Cli) delete(uri string) (*http.Response, error) {
+	method := "DELETE"
+	req, _ := http.NewRequest(method, uri, nil)
+	log.Info("%s %s", req.Method, req.URL.String())
+	if resp, err := c.makeRequest(req); err != nil {
+		return nil, err
+	} else {
+		if resp.StatusCode == 401 {
+			if err := c.CmdLogin(); err != nil {
+				return nil, err
+			}
+			req, _ = http.NewRequest(method, uri, nil)
+			return c.makeRequest(req)
+		}
+		return resp, err
+	}
+}
+
 func (c *Cli) makeRequestWithContent(method string, uri string, content string) (*http.Response, error) {
 	buffer := bytes.NewBufferString(content)
 	req, _ := http.NewRequest(method, uri, buffer)
--- a/commands.go
+++ b/commands.go
@@ -414,6 +414,62 @@ func (c *Cli) CmdWatch(issue string) error {
 	return nil
 }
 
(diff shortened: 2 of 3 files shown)


coryb:
Looks good.   In the interest of minimizing commands, perhaps we could minimize the usage to:
```bash
jira vote ISSUE [--down]
```
where vote `--down` == `unvote` and would call the DELETE.  I think that also fixes the DRY if we just pass a second arg to `CmdVote(issue string, up bool)` and the call would be something like `CmdVote(args[0], !opts["down"].(bool))`

Also I think you missed adding the command to the big usage text block.

I can fix these issues post merge if you like.
-Cory

mikepea:
Sounds good - actually works better for me this way: the func defines how
to treat that API entity rather than how to perform a specific action.

Similarly, looking at CmdWatch, there's no code to unwatch - which I need.

Would you be up for me changing the spec of that to be:

    func (c *Cli) CmdWatch(issue string, watcher string, up bool) error

... and push the population of 'watcher' up to main.go?

This works better for my needs for https://github.com/mikepea/go-jira-ui --
having to manipulate c.opts to call a Cmd*() is a pain (as I need to deal
with multiple actions in the same binary)

coryb:
yeah, sounds good. I guess for CmdWatch we should change the last param to "add"  (instead of "up"), so:
```
func (c *Cli) CmdWatch(issue string, watcher string, add bool) error
```

 with usage:
```
jira watch ISSUE [--watcher user] [--remove]
```
Then we call with:
`
CmdWatch(args[0], opts["user"].(string), !opts["remove"].(bool))
`

-Cory

coryb:
Well, i guess the lass call should be something like:
```go
watcher := c.GetOptString("watcher", opts["user"].(string))
CmdWatch(args[0], watcher, opts["remove"].(bool))
```
or somethine, where GetOptsString looks for options and returns the second (default) arg if not present.  Basically a getOrElse call.  The routine already exists in the `jira` package but I think it is private scope, so `main` wont have access, so we would need/want to make that public.

-Coy

mikepea:
Hey @coryb -- here's the `vote --down` update as discussed. I'll push up the CmdWatch update as a separate PR.

coryb:
Looks great, thanks!
-Cory

PR #27, opened 2016-01-21
Add insecure option for TLS endpoints

### LINKED ISSUE #25: add insecure option
For reasons I don't fully understand, yet, I'm unable to get Go to trust the CA that signed the certificate for my company's Jira installation.  Having an `--insecure-skip-verify` config option would be useful (and probably for other folks, too).

blalor:
This gives the option of disabling TLS certificate verification for the server.

Closes #25.

### DIFF
--- a/cli.go
+++ b/cli.go
@@ -11,6 +11,7 @@ import (
 	"net/http"
 	"net/http/cookiejar"
 	"net/url"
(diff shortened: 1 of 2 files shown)


coryb:
Nice, thanks for the patch!
-Cory

PR #29, opened 2016-01-22
[WIP] Add support for command aliases, stored queries

blalor:
Just opening this for discussion and feedback, right now.  I'd like to add support for saved queries (JQL) and command aliases.

`start <issue>` and `stop <issue>` could be defined in a config file as

``` yaml
aliases:
    start:
        description: transition an issue to "start"
        command: transition
        require_args: 1
        args:
            - {{ index 0 .args }}
            - start
    stop:
        description: transition an issue to "stop"
        command: transition
        require_args: 1
        args:
            - {{ index 0 .args }}
            - stop
```

and `jira list in-progress` as

``` yaml
queries:
    in-progress:
        description: shows issues in progress for the current user
        jql: >
            status in (Open, "In Progress", Reopened, "Feedback Required", Blocked)
            AND resolution = Unresolved
            AND assignee in (currentUser())
            ORDER BY updated DESC
```

I've started by refactoring the config to a `struct` from a `map[string]interface{}`, which should make for better readability and reduce errors.


coryb:
Hey, trying to grok what you have.

There is already support for custom queries (and custom templates) and command alias.  The are a bit non-conventional (perhaps crude) though as it is done by making the config file into a script. If `./.jira.d/config.yml` is executable, go-jira will execute it and parse the output.

So to create an in-progress command I would have:
```bash
#!/bin/bash
case $JIRA_OPERATION in
  in-progress)
    cat <<EOM
command: list
template: my-special-template
query: 'status in (Open, "In Progress", Reopened, "Feedback Required", Blocked) AND resolution = Unresolved AND assignee in (currentUser()) ORDER BY updated DESC'
EOM
    ;;
esac
```

which would be used like `jira in-progress`

Swapping out the opt map with the struct is definitely more the "go" way and probably the correct approach.  The commands-via-configuration is also interesting, I have been struggling trying to find a good cli library to make this easier.  I have looked at all the ones on awesome-go and they all fall short for me (I am very picky about the command line usage for my tools).  I have not fully evaluated kingpin, but it looked the most promising.

Let me think about this for a bit.  Thanks for the suggestions.
-Cory

blalor:
Have you looked at https://github.com/jessevdk/go-flags/?  I haven’t used it for sub-commands, but I really like its option parsing.

I’m not discounting the executable config support (I think that’s a rather brilliant design; really hackable) but for my own use I prefer the declarative approach, which I think can also be used to drive more user-friendly help, like showing all the stored queries and their descriptions.

I’m having trouble with the concept of overrides; it doesn’t seem to be well documented (or I missed it).  Maybe some more examples would help.  The way they’re currently implemented, my changes do break some of that.  I’m thinking of creating an Overrides map[string]interface{} field in the Options struct that would move all of those down a level.  The malleable config options are powerful, but ultimately harder to reason about in the code, in my opinion.

coryb:
Hey, sorry for the delay.  I spent the time converting one of my other projects to an approach similar to what you are recommending for go-jira.  I like it, so thumps-up.  The option parser optigo and go-flags look roughly equivalent, so not sure we would need to switch that. 

I also spent some times with kingpin which I actually quite like, but it cannot handle some of the usage go-jira has (ie: jira ISSUE blocks ISSUE).

Anyway, for Overrides, those are a catch-all option that are basically used to populate edit or create templates.  So in order to not require a specific command line option for every conceivable field that may be populated in the edit/create templates I put in the --override foo=bar usage.  So there are some duplicates like --assignee=someone where you could use --override assignee=someone, but in general the --override key=val is open ended.  Since users are allowed to create their own template we cannot predict all the various fields that would be used.

-Cory

blalor:
The things I really like about go-flags is that (a) automatically generates help (and man pages!) and (b) the way you specify the flags by way of tag data on the options struct, like [here](https://github.com/bluestatedigital/riemann-consul-receiver/blob/master/main.go#L21-L31).  I just find that means less code and maintenance.  I haven't played with its sub-command support, yet.

I finally got my head around overrides in https://github.com/blalor/go-jira/commit/603089aaa81e4f3fbe2daea6090526fca21f8b90.  I haven't touched custom commands, yet, but I'm *loving* the saved queries.  As far as custom commands go, I think there's a particular need for them because not everyone has the same transitions; I can't use `ack`, for example, and since it's hard-coded, I can't change it via config.

I'm still not sure how to go about adding custom commands, yet.  I'm currently chewing on the idea of defining the commands as strings that are processed with `text/template` and re-processed through the option parsing.  It's that or some kind of mini DSL as I hinted at initially for figuring out where options provided on the command line are placed.

coryb:
Curious if you have made any progress on command line alias via config.  It is interesting.

btw, I am open to replacing optigo with go-flags.  I would just hope that the command line interface does have to change.

blalor:
I haven’t had time to work on command aliases.  I’ll try to take a crack at replacing optigo with go-flags to see how that looks; I don’t think the flags will need to change.

gvol:
@blalor, @coryb, I would love to see this finished.  What do you guys think are the chances of that?  I don't know go, but if both of you have lost interest, then maybe I'll try to pick it up myself.

Right now, I am having some problems that (I believe) can't be solved with the executable config.  It works great for `jira mine` => `jira ls -q ...`, but what I want now is `jira add-to-sprint ISSUE` => jira `jira request -M POST "https://url/with/sprint" '{"issues":["ISSUE"]}'`.  But I don't have access to get or set arguments (as opposed to options) in config.yml.

coryb:
Hey @gvol, I am looking into adding alias support now.  I have rewritten go-jira on the `v1` branch and it is nearly stable.  I am looking to add something similar to what @blalor has proposed where it is driven via config files.

gvol:
Sweet!  In the mean time I have hacked together passing of arguments to and from config.yml and it works, but it feels really hacky to do that.  

On a side note, I can get the current sprint, but I can't actually convince Jira to move something into it.

rbuckland:
I hacked together the below this afternoon.

1. `mkdir ~/.jira.d`
2. `cp config.yml ~/.jira.d`
3. `chmod +x ~/.jira.d/config.yml`
4. `ln -s ~/.jira.d/config.yml ~/bin/j`
5. Edit `~/.jira.d/config.yml` and set your default project.

Now use `j` instead of `jira`  (not entirely perfect .. but I use `jira ... ` if I need to use it directly.
```
#!/bin/bash
#
# VERSION 0.1.0

if [ $(basename $0) == "j" ]; then

    case $1 in
        mine)   JIRA_CUSTOM_QUERY=mine  jira ls                      ;;
        epics)  JIRA_CUSTOM_QUERY=epics jira ls                      ;;
        search) JIRA_CUSTOM_QUERY=search JIRA_QUERY="${@:2}" jira ls ;;
        all)    JIRA_CUSTOM_QUERY=all   jira ls                      ;;
        *)      echo jira $*; jira "$@"                                              ;;
    esac


else

    #
    # Custom A&I jira Config
    #

    PROJECT=MYPRJ

    echo "endpoint: https://URL"
    echo "user: $USER"
    echo "password-source: keyring"
    echo "project: ${PROJECT}"
    echo "editor: $VISUAL"

    case $JIRA_CUSTOM_QUERY in
        mine)
            echo "query: assignee = currentUser() AND sprint in openSprints() ORDER BY rank ASC"
            ;;

        epics)
            echo "query: project = ${PROJECT} and issuetype = Epic and resolution is empty ORDER BY rank ASC"
            ;;

        search)
            echo "query: project = ${PROJECT} AND (resolution is EMPTY) AND  (summary ~ \"$JIRA_QUERY\" or comment ~ \"$JIRA_QUERY\" or description ~ \"$JIRA_QUERY\") ORDER BY rank ASC"
            ;;

        all)
            echo "query: project = ${PROJECT} ORDER BY rank ASC"
            ;;
        default)
            echo "query: project = ${PROJECT} AND (resolution is EMPTY) ORDER BY rank ASC"
            ;;
    esac

    case $JIRA_OPERATION in
        list)
            echo "template: table"
            ;;
    esac
fi

```

coryb:
Thanks @gvol and @rbuckland.  I have just released a major rewrite of `jira` and have implemented custom commands, you can read about them here: https://github.com/Netflix-Skunkworks/go-jira#custom-commands-1

I am going to close this issue.  I like the idea of custom queries that both @blalor and @rubuckland have mentioned, I have opened issue #107 to track that.

rbuckland:
Thanks @coryb nice. I will look at the current changes and #107 and see what I can make work.

rbuckland:
This might be a bit much :-) but @gvol , below is how you add tickets to the current sprint using the `go-jira`...

Follow thw `j add2sprint <issue>` trail.

```
#!/bin/bash
# Author: Ramon Buckland
# VERSION 0.4.0

PROJECT=MYPROJECT
PROJECT_BOARD="MySpecialProject%20-%20Scrum" # needs %20 for spaces in the board name

function current_sprint {

    local project_board="$1"

    board_id=$( jira request -M GET "/rest/agile/1.0/board?name=${PROJECT_BOARD}" | jq '.values[0].id' )
    sprint_id=$( jira request -M GET "/rest/agile/1.0/board/${board_id}/sprint?state=active" | jq '.values[0].id' )

    echo $sprint_id
}

function add_issue_to_current_sprint {

    local project_board="$1"
    local issue_id=$2

    sprint_id=$(current_sprint "$project_board")
    jira request -M POST  /rest/agile/1.0/sprint/${sprint_id}/issue "{\"issues\":[ \"${issue_id}\" ]}"
}


if [ $(basename $0) == "j" ]; then

    case $1 in
        # --custom for bash completion targets ~/bin/j is a symlink to the +x config.yml
        --custom) cat ~/bin/j | egrep ";;\s*#CUSTOM" | awk '{print $1}' | cut -d')' -f1 ;;
        # helper to show all these options
        mine)                   JIRA_CUSTOM_QUERY=mine                 jira ls ;;  #CUSTOM
        epics)                  JIRA_CUSTOM_QUERY=epics                jira ls ;;  #CUSTOM
        search)                 JIRA_CUSTOM_QUERY=search JIRA_QUERY="${@:2}" jira ls ;;  #CUSTOM
        search-all)             JIRA_CUSTOM_QUERY=search-all JIRA_QUERY="${@:2}" jira ls ;;  #CUSTOM
        all)                    JIRA_CUSTOM_QUERY=all                  jira ls ;;  #CUSTOM
        insprint)               JIRA_CUSTOM_QUERY=insprint             jira ls ;;   #CUSTOM
        insprint-closed)        JIRA_CUSTOM_QUERY=insprint-closed      jira ls ;;   #CUSTOM
        nosprint-open)          JIRA_CUSTOM_QUERY=nosprint-open        jira ls ;;   #CUSTOM
        nosprint-open-recent)   JIRA_CUSTOM_QUERY=nosprint-open-recent jira ls ;;   #CUSTOM
        add2sprint)      add_issue_to_current_sprint $AGILE_BOARD, $2     ;;   #CUSTOM

        *)      echo jira $*; jira "$@"                                              ;;
    esac


else
    #
    # Start of normal jira config.yml
    #


    echo "endpoint: https://my-jira-instance.com"
    echo "user: $USER"
    echo "password-source: keyring"
    echo "project: ${PROJECT}"
    echo "editor: $VISUAL"

    case $JIRA_CUSTOM_QUERY in
        mine)
            echo "query: assignee = currentUser() AND sprint in openSprints() ORDER BY rank ASC"
            ;;

        epics)
            echo "query: project = ${PROJECT} and issuetype = Epic and resolution is empty ORDER BY rank ASC"
            ;;

        search-all)
            echo "query: project = ${PROJECT} AND  (summary ~ \"$JIRA_QUERY\" or comment ~ \"$JIRA_QUERY\" or description ~ \"$JIRA_QUERY\") ORDER BY rank ASC"
            ;;

        search)
            echo "query: project = ${PROJECT} AND (resolution is EMPTY) AND  (summary ~ \"$JIRA_QUERY\" or comment ~ \"$JIRA_QUERY\" or description ~ \"$JIRA_QUERY\") ORDER BY rank ASC"
            ;;

        all)
            echo "query: project = ${PROJECT} ORDER BY rank ASC"
            ;;

        insprint)
            echo "query: project = ${PROJECT} and sprint in openSprints() AND resolution is EMPTY ORDER BY rank ASC"
            ;;

        insprint-closed)
            echo "query: project = ${PROJECT} and sprint in openSprints() AND resolution is not EMPTY ORDER BY rank ASC"
            ;;

        nosprint-open)
            echo "query: project = ${PROJECT} and ( sprint not in openSprints() OR sprint is EMPTY ) AND resolution is EMPTY ORDER BY rank ASC"
            ;;

        nosprint-open-recent)
            prev_date=$( date -v-3w  +%Y-%m-%d )
            echo "query: project = ${PROJECT} and ( sprint not in openSprints() OR sprint is EMPTY ) AND created > ${prev_date} AND resolution is EMPTY ORDER BY rank ASC"
            ;;

        default)
            echo "query: project = ${PROJECT} AND (resolution is EMPTY) ORDER BY rank ASC"
            ;;
    esac

    case $JIRA_OPERATION in
        list)
            # Change the output format with this line (and use the jira templates)
            # echo "template: table"
            # I like just a list for "ls" entries .. tables was too busy. See ~/.jira.d/templates/list for fancy example
            echo "queryfields: summary,created,updated,priority,status,reporter,assignee,labels,issuetype"
            ;;
    esac
fi
```

PR #30, opened 2016-01-24
Support for removing a given watcher

mikepea:
As per discussion in #26 -- here is 'remove watcher' support, eg:

    jira watch EXAMPLE-123 --watcher bob --remove

I've exposed GetOptString and GetOptBool in the same manner as before - leave the private function intact, just add a public wrapper.

I've mentioned in the commit, but I've expressly not included 'remove: false' as a default option, since at the moment this is specific to 'watch' command. Maybe if we find other reasons to add '--remove' we can promote it.

NOTE: this commit involves a breaking change to the CmdWatch contract, in that it now takes the watcher and whether to remove as arguments. I think i'm the only consumer of this as a library right now, so this is likely not a big problem.

### DIFF
--- a/cli.go
+++ b/cli.go
@@ -474,12 +474,20 @@ func (c *Cli) FindIssues() (interface{}, error) {
 	}
 }
 
+func (c *Cli) GetOptString(optName string, dflt string) string {
+	return c.getOptString(optName, dflt)
+}
+
 func (c *Cli) getOptString(optName string, dflt string) string {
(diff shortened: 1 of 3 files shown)


mikepea:
Oh, @coryb  -- this is based off the `vote_support` PR in #26 (for DELETE support), so that will need to be merged first.

coryb:
Looks good, thanks again!
-Cory

PR #33, opened 2016-01-29
Fixes #32 - make path to cookieFile if it's not present

### LINKED ISSUE #32: Saving cookies fails if ~/.jira.d is not present
If `/etc/go-jira.yml` is used to set the endpoint, then there is no reason for the user to create their ~/.jira.d directory manually.

This then leads to CmdLogin breaking, as it cannot save the cookies file:

    2016-01-29T11:02:54.109Z ERROR [util.go:252] Failed to open /Users/mikepea/.jira.d/cookies.js: open /Users/mikepea/.jira.d/cookies.js: no such file or directory

mikepea:
TL;DR, this ensures ~/jira.d is present, with 0755 perms.

If `~/jira.d` isn't present, we can't write to the cookieFile, which
breaks CmdLogin. This is particularly an issue when using `/etc/go-jira.yml`
to get an entire team using go-jira easily :)

This fixes this by ensuring the cookieFile dir is present before
writing to it.

### DIFF
--- a/cli.go
+++ b/cli.go
@@ -14,6 +14,7 @@ import (
 	"net/url"
 	"os"
 	"os/exec"
+	"path"
 	"runtime"
 	"strings"
 	"time"
@@ -87,6 +88,7 @@ func (c *Cli) saveCookies(cookies []*http.Cookie) {
 		}
 		jsonWrite(c.cookieFile, mergedCookies)
 	} else {
+		mkdir(path.Dir(c.cookieFile))
 		jsonWrite(c.cookieFile, cookies)
 	}
 }


mikepea:
Hey @coryb -- this fixes a blocker for me announcing go-jira and go-jira-ui to the wider engineering team here, could you merge and cut a new release?

I guess the release should be 0.1.0 due to the change to the CmdWatch interface aye?

coryb:
Okay, I will bump the version and get a release out.  Thanks

-Cory

coryb:
Created the new release: 
https://github.com/Netflix-Skunkworks/go-jira/releases/tag/v0.1.0

With the vX.Y.Z tags you should be able to use gopkg.in instead of github for your import if you want to be more insulated from go-jira changes, so I think this will work:
`gopkg.in/Netflix-Skunkworks/go-jira.v0.1`

-Cory

PR #34, opened 2016-02-01
Fix issuetype calls adding URL escaping

jonathanio:
It is valid within JIRA to have issue types with spaces (for example we use "Pro-Active Task") however the `issuetype` variable is not escaped prior to formatting into the `uri` string.

This fix imports `net/url` and escapes all inclusions of `issuetype` into `uri`. The only other variables formatted into `uri` are `c.endpoint` and `issue` which shouldn't contain special characters.

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -7,6 +7,7 @@ import (
 	"golang.org/x/crypto/ssh/terminal"
 	"net/http"
 	"net/http/httputil"
+	"net/url"
 	"os"
 	"strings"
 	// "github.com/kr/pretty"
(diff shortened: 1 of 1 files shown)


jonathanio:
As an example:

```
2016-02-01T10:49:16.036Z DEBUG [cli.go:175] GET /rest/api/2/issue/createmeta?projectKeys=TEST&issuetypeNames=Management Task&expand=projects.issuetypes.fields HTTP/1.1
Host: example.atlassian.net
User-Agent: Go-http-client/1.1

2016-02-01T10:49:16.729Z ERROR [commands.go:249] Project 'TEST' does not support issuetype 'Management Task'.  Unable to create issue.
```

coryb:
Looks good, thanks!
-Cory

PR #35, opened 2016-02-02
Fix path for installation instructions

QuinnyPig:
go-jira/jira isn't a proper path.


coryb:
Whoops, yeah, moved code around a few weeks back. Thanks!
-Cory

PR #37, opened 2016-02-22
Added support for the ```expand``` option for Issues

tobyjoe:
The ```expand``` option is used to specify resource expansion in the
Jira REST API.

It's particularly useful for things like fetching the ```changelog``` of
an Issue.

This PR adds the support to the ```ListIssues``` and ```ViewIssue```
functions of the ```jira.Cli``` struct.

I'm happy to add tests, but there is currently no test suite in the
master branch, so I did not want to bog down the PR with tangential features.

### DIFF
--- a/cli.go
+++ b/cli.go
@@ -21,7 +21,7 @@ import (
 )
 
 var (
-	log     = logging.MustGetLogger("jira")
+	log = logging.MustGetLogger("jira")
 	VERSION string
 )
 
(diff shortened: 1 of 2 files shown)


cli.go
```diff
@@ -402,6 +402,10 @@ func (c *Cli) SaveData(data interface{}) error {
 
 func (c *Cli) ViewIssue(issue string) (interface{}, error) {
 	uri := fmt.Sprintf("%s/rest/api/2/issue/%s", c.endpoint, issue)
+	if x := c.expansions(); len(x) > 0 {
```
tobyjoe: This would be more efficient with ```fmt.Sprintf``` only called once, and the right one being selected per branch of ```if... else...``` but I found it easier to read this way - especially for an unsolicited PR. And even though ```fmt.Sprintf``` isn't super cheap, this tool is not really gonna be impacted by an extra operation or two for most users :snail:

coryb:
Looks good, thanks!
-Cory

PR #39, opened 2016-03-30
Include templates from a system path

mikepea:
I've found that several of the built-in templates need a bit of work
to be useful with our JIRA installation (eg #24), so this allows for admins
to make a site-specific set of defaults easily.

### DIFF
--- a/cli.go
+++ b/cli.go
@@ -213,33 +213,33 @@ func (c *Cli) GetTemplate(name string) string {
 	return c.getTemplate(name)
 }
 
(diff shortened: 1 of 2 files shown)


coryb:
Thanks for the PR.  I am also open to modifying the default templates to make them more generically compatible with "stock" jira installs (if there is such a thing).

-Cory

mikepea:
Thanks Cory - i'll submit some updates to the default templates separately, but aiui the 'watchers' field at least varies pretty wildly between deployments, so I doubt we're going to find a truly generic default :(

PR #49, opened 2016-08-04
Added initial workload commands

RC1140:
Havent added tests yet , will add them shortly , this is just so that the team can get an initial overview of the work done and suggest changes.

### DIFF
--- a/cli.go
+++ b/cli.go
@@ -440,6 +440,42 @@ func (c *Cli) SaveData(data interface{}) error {
 	return nil
 }
 
+func (c *Cli) NewIssueWorkLog(issue string, timeSpentInSecond int, comment, started string) (interface{}, error) {
+	uri := fmt.Sprintf("%s/rest/api/2/issue/%s/worklog", c.endpoint, issue)
+	if x := c.expansions(); len(x) > 0 {
+		uri = fmt.Sprintf("%s?expand=%s", uri, strings.Join(x, ","))
+	}
+
+	json, err := jsonEncode(map[string]interface{}{
+		"timeSpentSeconds": timeSpentInSecond,
+		"comment":          comment,
+		"started":          started,
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	if data, err := responseToJson(c.post(uri, json)); err != nil {
+		return nil, err
+	} else {
+		return data, nil
+	}
+}
+
+func (c *Cli) ViewIssueWorkLogs(issue string) (interface{}, error) {
+	uri := fmt.Sprintf("%s/rest/api/2/issue/%s/worklog", c.endpoint, issue)
+	if x := c.expansions(); len(x) > 0 {
+		uri = fmt.Sprintf("%s?expand=%s", uri, strings.Join(x, ","))
+	}
+
+	data, err := responseToJson(c.get(uri))
+	if err != nil {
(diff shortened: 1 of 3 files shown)


coryb:
Overall it looks good, thanks!

Some thoughts, your usage seems to require a 4th argument "started"
```
jira addworklog ISSUE time comment started
```

Although this interface seems a bit cumbersome.  I wonder if this should not be an editable template like we use for the `create` or `edit` commands?  Something like:
```
$ jira add worklog ISSUE
>> editor
comment: work hard
timeSpent: 3h 20m
started:
```
Then if `started` is blank just convent `timeSpent` and assume `started = now - timeSpent`?  (not sure with Jira if we can just pass the `timeSpent`, or if we are required to send the `timeSpentSeconds` field as well)

Then users that dont wish to interactively edit could do something like:
```
jira add worklog ISSUE -m "work hard" -o timeSpent="3h 20m" --noedit
```

I also prefer the "add worklog" subcommand style vs the single "addworklog" command.  The subcommand style matches the "add component" and "add labels" command syntax.  With this syntax it is a bit fussier to implement though, both of the new commands will appear as "worklog"  but "add worklog" will have "add" as `args[0]` via this [funny fixup code](https://github.com/Netflix-Skunkworks/go-jira/blob/master/main/main.go#L248).

For the worklog command, you can cheat for the template and just set the "worklog" template to the debug template [here](https://github.com/Netflix-Skunkworks/go-jira/blob/master/templates.go#L3) then the code you have in TODO will mostly work after replacing "view" with "worklog".  Then we can work on getting a pretty template eventually.

RC1140:
Cool , will update it this evening and let you know when its done.

coryb:
btw, dont worry about writing real tests, there are none in go-jira yet.  I am working on creating an integration test suite [here](https://github.com/Netflix-Skunkworks/go-jira/tree/master/t), but it is not really ready for general consumption yet, so it would be unfair to ask for that.  I will add tests for this once I get that stuff working.
-Cory

coryb:
Well, I merged this work, but for some reason github does not think I merged it.  There were some conflicts due to some refactoring I did recently (golint etc) and I didnt want to make you suffer through the merge conflicts.    Then I tweaked the api based on our previous discussion adding a "worklogs" template for the `jira worklog ISSUE` command, and adding a "worklog" template for the `jira add worklog ISSUE` command.  I didnt want to "step on your toes" but you seem pretty busy and the work was mostly done, so I thought I could help out and finish it.

So now `jira worklog ISSUE` will work as well as something like `jira add worklog ISSUE -m "work hard" -T "3h 20m" --noedit`  where -T is short for `--time-spent`.  I removed the `started` values since Jira does not seem to require that, but you can add it directly into the template when edit via `jira add worklog ISSUE`.

So thanks a lot for the help.  I am sorry your commits dont appear in the git log though, not sure what I did wrong to make that happen, I just followed the "resolve conflicts" workflow as documented by github.

Btw, I also added a quick test for this:
https://github.com/Netflix-Skunkworks/go-jira/blob/master/t/110basic-worklog.t

Anyway, thanks a lot.
-Cory

RC1140:
Thanks for getting this in , I do apologize for the delays in your follow up, but glad you got it sorted in the end

PR #52, opened 2016-08-24
Prefer transition names which match exactly

dbrower:
Some transitions are substrings of other transitions, and the current loop sometimes doesn't choose the best match. For example, it would never choose "QA" if there were also a transition "Deploy to QA". Ditto for "Open" and "Reopen".


coryb:
Looks good, thanks!
-Cory

coryb:
Released:
https://github.com/Netflix-Skunkworks/go-jira/releases/tag/v0.1.7

PR #53, opened 2016-08-25
Add support for a unix socket proxy

jshirley:
Hi there!

### Summary

We use client certificate authentication, and have a running local proxy on our laptops that allows us to proxy to our internal services, such as JIRA. This is accomplished by shadowing the transport with a custom dialer. It works on my machine :sparkles:. I'm inexperienced with Go a bit, so please advise if I'm doing something absurd!

### What's this PR do?

1. Adds a unixproxy.go file which implements a transport with a shadow to use a unix socket
2. Adds a `unixproxy` configuration option to point to an path on the computer to use
3. If configured, uses the unix proxy as a transport for all http requests

Thanks!

### DIFF
--- a/cli.go
+++ b/cli.go
@@ -43,26 +43,35 @@ func New(opts map[string]interface{}) *Cli {
 	endpoint, _ := opts["endpoint"].(string)
 	url, _ := url.Parse(strings.TrimRight(endpoint, "/"))
 
-	transport := &http.Transport{
-		TLSClientConfig: &tls.Config{},
-	}
-
 	if project, ok := opts["project"].(string); ok {
 		opts["project"] = strings.ToUpper(project)
 	}
 
-	if insecureSkipVerify, ok := opts["insecure"].(bool); ok {
-		transport.TLSClientConfig.InsecureSkipVerify = insecureSkipVerify
+	var ua *http.Client
+	if unixProxyPath, ok := opts["unixproxy"].(string); ok {
+		ua = &http.Client{
+			Jar:       cookieJar,
+			Transport: UnixProxy(unixProxyPath),
+		}
+	} else {
+		transport := &http.Transport{
+			TLSClientConfig: &tls.Config{},
+		}
+		if insecureSkipVerify, ok := opts["insecure"].(bool); ok {
+			transport.TLSClientConfig.InsecureSkipVerify = insecureSkipVerify
+		}
+
+		ua = &http.Client{
+			Jar:       cookieJar,
+			Transport: transport,
+		}
 	}
 
 	cli := &Cli{
 		endpoint:   url,
 		opts:       opts,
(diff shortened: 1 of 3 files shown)


cli.go
```diff
@@ -43,26 +43,35 @@ func New(opts map[string]interface{}) *Cli {
...
-	}
-
 	if project, ok := opts["project"].(string); ok {
 		opts["project"] = strings.ToUpper(project)
 	}
 
-	if insecureSkipVerify, ok := opts["insecure"].(bool); ok {
-		transport.TLSClientConfig.InsecureSkipVerify = insecureSkipVerify
+	var ua *http.Client
+	if unixProxyPath, ok := opts["unixproxy"].(string); ok {
```
coryb: the main/main.go will need to be modified to allow for this option I think (look for `optigo.NewDirectAssignParser`), and the usage should be updated to list the option and specify format.  (I was not sure if this should be just a path, or if it should have a `unix://` prefix ... maybe should allow for both?)
-Cory

coryb:
Looks interesting, update the usage and I will merge.

Not sure how we can test this easily.  I imagine `socat` could be used to setup a socket/http proxy for testing this, but not sure.  If you have any ideas on how I can simulate your setup  I can play around with it.

Thanks!
-Cory

coryb:
@jshirley any updates on my comments?
Thanks
-Cory

jshirley:
Eep, really sorry! This fell off my plate. I'll revise this coming week. 






On Sun, Sep 18, 2016 at 12:00 AM -0700, "coryb" <notifications@github.com> wrote:












@jshirley any updates on my comments?

Thanks

-Cory

jshirley:
Hey Cory,

I updated (again, sorry for the delay on this) the usage. Testing without a real client cert proxy is a bit tough, but I'm using it locally here with our client cert adapter thingy and it's working great.
```
$ jira ls -p DEVTOOLS
DEVTOOLS-567: <redacted>
...
```

Let me know if there is anything specific you'd like me to test out, too. Our certproxy isn't public because it has some details about the shape of our client certs but we could potentially clean that up and make it public (just haven't had much of a reason to yet).

coryb:
Okay, I will merge this, still not clear on the format of the --unixproxy option though  Is it just a path to socket file or does it require a `unix://` scheme.

-Cory

jshirley:
Ah, I'll improve that!






On Thu, Sep 22, 2016 at 6:20 PM -0700, "coryb" <notifications@github.com> wrote:












Okay, I will merge this, still not clear on the format of the --unixproxy option though  Is it just a path to socket file or does it require a unix:// scheme.



-Cory

coryb:
Looks good, thanks.  The test failure is unrelated, I will debug that later. 

-Cory

PR #61, opened 2016-12-17
fix(http): Add proxy transport

sylus:
Hi there,

First of all thanks so much for this great tool!

I did notice while working with go-jira I was unable to get it to work with my corporate proxy but am not sure if I am missing something

I added support just for checking if HTTP_PROXY is set, although I thought this would natively be handled by go.

Let me know if this is the correct approach! thx!

### DIFF
--- a/cli.go
+++ b/cli.go
@@ -63,6 +63,11 @@ func New(opts map[string]interface{}) *Cli {
 			transport.TLSClientConfig.InsecureSkipVerify = insecureSkipVerify
 		}
 
+		if os.Getenv("HTTP_PROXY") != "" {
+			proxyURL, _ := url.Parse(os.Getenv("HTTP_PROXY"))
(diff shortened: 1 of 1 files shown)


coryb:
Hey, looked into this a bit, I think we just need to set the Proxy parameter on the http.Transport object, like this:

```
        transport := &http.Transport{
            Proxy: http.ProxyFromEnvironment,
            TLSClientConfig: &tls.Config{},
        }
```

That should fix your issue.  You want to try this out and update your patch?

Btw, dont worry about the test failure, I need to fix that.

Thanks!
-Cory

sylus:
That indeed did the trick! Thanks so much!

I have updated my P.R. ^_^

coryb:
Excellent, thanks!  I will get a release out soon with this.

-Cory

PR #78, opened 2017-04-24
Added generic issuelink command

davidreuss:
This allows adding generic links, and could replace 'blocks', and 'dups'
command, since it's pretty much just a copy/paste job.

Usage will be something like:

`$ jira issuelink $INWARDISSUE "Relates" OUTWARDISSUE`

Pulling the list of the names, for your issuelinktypes

```
$ jira issuelinktypes | jq '.issueLinkTypes | map(.name)'
[
  "Blocks",
  "Bonfire testing",
  "Clones",
  "Deprecates",
  "Duplicate",
  "Relates",
  "Risks"
]
```

### DIFF
--- a/commands.go
+++ b/commands.go
@@ -531,6 +531,50 @@ func (c *Cli) CmdIssueLinkTypes() error {
 	return runTemplate(c.getTemplate("issuelinktypes"), data, nil)
 }
 
+// CmdIssueLink is a generic function for adding a link type to an issue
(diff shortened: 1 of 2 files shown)


davidreuss:
Duh, just saw this ..

https://github.com/Netflix-Skunkworks/go-jira/issues/72#issuecomment-285921538

So, it could be overridden, it's just .. uh, weird UX, since the command is still called either `blocks` or `dups`.

coryb:
Adding a generic command makes sense to me, thanks this patch!
-Cory

PR #90, opened 2017-07-11
Use name field for fix version name value. Fixes #89

### LINKED ISSUE #89: Transitioning an issue can fail with map fix version value
Tried to transition an issue with a fix version assigned and it failed with Bad Request,

```
{"errorMessages":[],"errors":{"fixVersions":"Version name 'map[released:false self:https://jira.company.com/rest/api/2/version/25726 id:25726 name:component-0.6.0 archived:false]' is not valid"}}
```

When I choose to edit during the transition I see the fix version value is set to,

```
fields:
  fixVersions: 
    - name: map[self:https://jira.company.com/rest/api/2/version/25726 id:25726 name:component-0.6.0 archived:false released:false]
```

Changing the `name` value to `component-0.6.0` works.

bbaugher:
I tried this out myself and it fixed it for me


coryb:
nice, looks good. Thanks!  

The tests are broken at the moment, I will try to fix them in a bit.

-Cory

PR #91, opened 2017-07-14
add documentation to auth with pass

kBite:
Currently README.md is missing a crucial part: when I configure go-jira to use pass, which entry is it looking for? It's 'GoJira/<user>'. Documentation to create this entry was added to README.md.

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,6 +222,12 @@ Then initialize the `pass` tool to use the correct key:
 $ pass init "Go Jira <gojira@example.com>"
 ```
 
+`go-jira` expects user's password to be stored under `GoJira/<user>`:
+```
+$ pass insert GoJira/<user>
+```
+
+
 You probably want to setup gpg-agent so that you dont have to type in your gpg passphrase all the time.  You can get `gpg-agent` to automatically start by adding something like this to your `$HOME/.bashrc`
 ```bash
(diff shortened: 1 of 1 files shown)


coryb:
Hey thanks, I am confused though.  go-jira will insert the password to the `pass` store the first time it prompts you if it is missing: https://github.com/Netflix-Skunkworks/go-jira/blob/master/password.go#L65  So I am surprised to see that you thought it was required to manually register your password with `pass`.  Did it not work for you that way?

kBite:
No, it didn't. I can reproduce this:
```
# rm manually set pw
$ pass rm -r GoJira
$ jira fields
Jira Password [user]: ********
$ echo $?
0
$ pass show | grep -i GoJira; echo $?
1
```
Configration is correct as I can use go-jira with pass after manually adding my account to pass.

I never debugged go code, but maybe I can give you information to fix this.

coryb:
Interesting, thanks for the confirmation.  That line section of code you indicate is just checking to see if the `password-source` has any value, if so it calls `SetPass`, so I would think it would work in your case.  I will try to reproduce on my end and see if I can get to the bottom of it.
-Cory

kBite:
An issue seems to me more appropriate to track this bug (#97).

PR #104, opened 2017-08-21
Handle keyring.ErrNotFound error instead of panicing.

wrouesnel:
github.com/tmc/keyring returns keyring.ErrNotFound instead of a blank password in newer libraries. Handle this case properly by nulling the error, to give the user a chance to set a new password.

This makes go-jira compatible with the newer versions which now incorporate a fix for working out-of-the-box on Linux.

### DIFF
--- a/keyring.go
+++ b/keyring.go
@@ -5,7 +5,11 @@ package jira
 import "github.com/tmc/keyring"
 
 func keyringGet(user string) (string, error) {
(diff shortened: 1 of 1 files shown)


coryb:
Nice, thanks!

coryb:
Released in 0.1.15
https://github.com/Netflix-Skunkworks/go-jira/releases/latest

PR #114, opened 2017-09-22
Go 1.9 fixes a problem with SSL certificate discovery on Mac OS.

wrouesnel:
Prior to 1.9 only the deepest level of system SSL certs were used by the Golang
SSL library, which meant that users with custom certificates would not have
them discovered (unless they edited the much lower level store). This is fixed
in Go 1.9, and thus upping the release version will fix it for all users.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ before_install:
 language: go
 go_import_path: gopkg.in/Netflix-Skunkworks/go-jira.v1
 go:
-  - 1.8
+  - 1.9
 
 matrix:
   fast_finish: true


coryb:
Hey, thanks.  That files just effects the test execution environment.  The binaries attached to the release are build using [xgo](github.com/karalabe/xgo) which apparently does not support go 1.9 yet.  This is how we build the binaries:
https://github.com/Netflix-Skunkworks/go-jira/blob/master/Makefile#L40

I tried modifying that to use `xgo --go 1.9` but I get:
```
Checking for required docker image karalabe/xgo-1.9... not found!
```
I will file an issue there, hopefully it will be easy to get a xgo 1.9 version.

wrouesnel:
Yeah I realised right as I pressed submit (and then didn't find a better fix before you got to it). Thanks for following up!

coryb:
I filed a few upstream PR's but they have not been dealt with, not sure if that project is still actively maintained.  For now I have created the docker images locally and used them to rebuild the binaries on the latest release:
https://github.com/Netflix-Skunkworks/go-jira/releases/tag/v1.0.10

Could you verify they work with your custom certs? 

Thanks
-Cory

wrouesnel:
I tried that release link on my Macbook and it didn't work - but I think it would have to be that whatever got published isn't a go 1.9 binary (for comparison a local build with go 1.9 did work for me).

If you don't mind me refactoring to mess with the build system a bit I can pick up figuring out how to get this done in this PR.

coryb:
Not sure how it would not be built with 1.9.  I just jumped in the container and it looks good:
```
$ docker run -ti --entrypoint "/bin/bash" karalabe/xgo-1.9.0
root@62b0d0833568:/# go version
go version go1.9 linux/amd64
```

In theory I am okay with refactoring the build, but xgo is the most elegant solution I have found so far.  There are cross platform build issues that xgo solves by providing native system libraries to compile against (like issues with os/users.Current)
-Cory

coryb:
Is this still an issue?  We used go-1.10 for the latest release:
```
$ docker run -ti --rm --entrypoint go karalabe/xgo-latest version
go version go1.10 linux/amd64
```

Issue #121, opened 2017-10-09
Unable to override resolution


gvol:
No matter what I do, `.overrides.resolution` seems to always be Done.  I tried passing`-o resolution=XXX` and `-o defaultResolution=XXX`, but to no avail.  It seems that around https://github.com/Netflix-Skunkworks/go-jira/blob/master/jiracmd/transition.go#L112 a `--resolution` flag is queried, but trying to use that with a transition command gives an error.  Then when it's not found it goes to the default which turns out to be "Done" for the Jira instance I'm using.

coryb:
I am not sure about this one.  Jira requires a resolution, so I know we try to pick a reasonable default, but you should be able to explicitly define it.  I will try to figure out what is wrong.

gvol:
@coryb, Does this PR look okay?

coryb:
Hey, sorry missed your update, it has been merged.  Thanks!

PR #124, opened 2017-10-12
Allow overriding resolution

gvol:
I'm _not_ advocating that this be the way to solve #121, but it works for me and sometimes it's helpful to have a patch, even a bad one.  :-)

### DIFF
--- a/jiracmd/transition.go
+++ b/jiracmd/transition.go
@@ -63,6 +63,7 @@ func CmdTransitionUsage(cmd *kingpin.CmdClause, opts *TransitionOptions) error {
 		cmd.Arg("TRANSITION", "State to transition issue to").Required().StringVar(&opts.Transition)
 	}
 	cmd.Arg("ISSUE", "issue to transition").Required().StringVar(&opts.Issue)
+	cmd.Flag("resolution", "Set resolution on transition").StringVar(&opts.Resolution)
 	return nil
 }
 


coryb:
Thanks for this, I am swamped now and really appreciate the help.  I think your solution is reasonable, but I wonder if it would just make more sense to expose a `--resolution` option instead?  That would preserve the existing logic, and require just adding this to the usage function:
```
cmd.Flag("resolution", "Set resolution on transition").StringVar(&opts.Resolution)
```

gvol:
Sounds like a great idea.  I tested and it works (after I reverted my other changes).

coryb:

Maratyz:
The resolution will be updated to "Done" only with state of "Finish" after I tried with a couple of transition state ("Start", "Reopen", "Wait for Response", "Pause"), e.g. `jira trans --resolution="Done" "Wait for Response" --noedit JIRA-123`. 

Is this intended design? 

I also tried `override` in this way `jira edit -o resolution="Done" --noedit JIRA-123`, and it does not work neither.

coryb:
@Maratyz i am not a Jira admin so I could be wrong, but it sounds like your Jira admins have set up some non-standard workflows.  The Jira api will abide by the rules set up for your project by the workflows, and there is nothing that I know of to allow `go-jira` to bypass those workflows.  I think you can only set a resolution on a "final" state for the workflows, sounds like "Finish" is a final workflow state for your project.

Maratyz:
HI @coryb , thanks for your reply, which is helpful for me to check further.

PR #126, opened 2017-10-19
Add regexReplace template function

schorsch3000:
#### What's this PR do?
It add a regRepace function to the template functions.
#### How should this be manually tested?
Have a template like `{{ .fields.description | regReplace "foo" "bar" }}`
#### Any background context you want to provide?
Alot of out Tickets are created via email.
Jira often creates double newlines, one for \n and one for \r, also People tend to use a lot of newlines in Email.
This results in tickets, larger than one full screen terminal, containing mostly newlines.
To solve this, and have simple but powerful code i decided to implement regexReplace instead of this specific behavior.

### DIFF
--- a/jiracli/templates.go
+++ b/jiracli/templates.go
@@ -8,6 +8,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"regexp"
 	"reflect"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


coryb:
Hey, thanks for the patch!  Sorry for the delay.

coryb:

PR #127, opened 2017-10-25
Create Metadata Not Populated Correctly

### LINKED ISSUE #123: meta data is not populated for `jira create`
I see the full json (with fields and allowedValues) coming back from the `jira createmeta` request (with `-v -v -v`, but it doesn't translate into the output.  Looking at it a little, it seems that jiradata/IssueType.go doesn't have Fields defined.  Do I need to update the schema somehow?  Maybe running schemas/fetch-schemas.py against my instance of Jira?  I can send more debug information if desired.

thedillonb:
Fixes #123

### DIFF
--- a/jiradata/IssueType.go
+++ b/jiradata/IssueType.go
@@ -46,11 +46,12 @@ package jiradata
 //   }
 // }
 type IssueType struct {
-	AvatarID    int    `json:"avatarId,omitempty" yaml:"avatarId,omitempty"`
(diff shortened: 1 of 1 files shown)


coryb:
Thanks for the patch, I have added in a test to make sure that the `Fields` field does not accidentally disappear if we have to regenerate the files again.

coryb:

PR #128, opened 2017-10-26
Add URL escaping to an additional issuetype call

mivok:
In #34 a fix was made to escape the issue type to support issue types
with spaces in. This is an additional place where the issue type needs
escaping, otherwise you get 400 Bad Request when trying to create an
issue for an issuetype that has a space in.


coryb:
Nice, thanks!
-Cory

coryb:
released https://github.com/Netflix-Skunkworks/go-jira/releases/tag/v1.0.13

PR #130, opened 2017-11-02
add support for using socks proxy

onionjake:
Adds support for using a socks5 proxy.


coryb:
Nice, thanks for the patch!

PR #142, opened 2018-01-19
Update unexportTemplates.go

anthonyrisinger:
Logic error, `-d` is set but not used.


coryb:
Doh, good find, thanks!
-Cory

PR #150, opened 2018-03-07
make it easier to compile with aribtrary version of go via GO env var

catskul:
Ubuntu 16.04 has go PPAs for the latest version, but does not link them to /usr/bin/go. As a result to use any other version of go than the default requires full path to go.

Per the standards used for c and c++ this adds a $GO variable which defaults to `go` which can be passed in via env var for a specific version of go.

Example:
```
export GO=/usr/lib/go-1.9/bin/go
$ make
/usr/lib/go-1.9/bin/go build -gcflags="-e" -v -ldflags "-w -s" -o '/home/asomerville/go/src/gopkg.in/Netflix-Skunkworks/go-jira.v1/jira' cmd/jira/main.go
command-line-arguments
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,5 @@
 NAME=jira
+GO?=go
 
 OS=$(shell uname -s)
 ifeq ($(filter CYGWIN%,$(OS)),$(OS))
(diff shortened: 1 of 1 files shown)


catskul:
This may be unnecessary since/if the makefile is only for travis' sake

coryb:
Most people setup the PATH to set preferred binaries, ie:
```
export PATH=/usr/lib/go-1.9/bin:$PATH
```

So I am not sure this patch is necessary.

catskul:
Feel free to close.

I'm mostly a c++ developer so I was matching idioms I'm used to w.r.t. gcc/clang version + Makefiles

coryb:
It is a small change that wont break anything, so merging in.  Thanks.

PR #153, opened 2018-03-08
Document shell completion

catskul:
Add in a stanza to the Usage section about tab completion

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@
          * [keyring password source](#keyring-password-source)
          * [pass password source](#pass-password-source)
    * [Usage](#usage)
-       * [TAB completion](#tab-completion)
+       * [TAB completion](#setting-up-tab-completion)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -517,3 +518,16 @@ Commands:
   worklog list:        Prints the worklog data for given issue
 
 ```
+
+#### Setting up TAB completion
```
coryb: This wont match your #tab-completion anchor.  GH Markdown is not that clever.  So you can change the TOC to `#setting-up-tab-completion` or change the topic to `#### TAB completion`.

catskul: Good catch. Fixed!

coryb:
Thanks for the help!

catskul:
No problem. So glad to have a usable tool that keeps me away from the browser. Thanks for making this tool!

PR #171, opened 2018-04-27
Add config possibility to set pass path

dvogt23:
hey,

have some small changes in config possibilities in the usage with pass.

greet

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 1.0.17 - 2018-04-15
 
+* fix IsTerminal usage for windows [Cory Bennett] [[7f9595c](https://github.com/Netflix-Skunkworks/go-jira/commit/7f9595c)]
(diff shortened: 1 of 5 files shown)


coryb:
Hey @dvogt23, sorry I missed this PR.  Could you explain the use-case for this?  I was wondering if it made more sense to completely override the password path, rather than just the `GoJira` prefix?  Maybe something like this:
```
if o.PasswordSource.Value == "pass" {
	if o.PasswordPath.Value != "" {
		return o.PasswordPath.Value
	}
	return fmt.Sprintf("GoJira/%s", user)
}
```

I am wondiner if the `PasswordPath` variable should be renamed, perhaps something like `PasswordName`?  I am concerned that `PasswordPath` would be confusing with `PasswordDirectory` since "path" and "directory" are often synonyms.

dvogt23:
Hey @coryb thanks for feedback and this good suggestions. 

In my use case, i have already an entry for the jira login and in my opinion, fixed paths/names can be better optionally moved to a config.

So who should make those changes? I could change my commit, or?

coryb:
I have updated the code and merged it in. 
Using `password-name` property, and it will be used to completely override the password entry.   So the default value would be something like:
```
password-name: GoJira/username@example.com
```

PR #180, opened 2018-05-23
Add Docker support

pataquets:
Add Dockerfile to enable image building. Useful for development and tests under Docker workflow.
Using the official Golang image, latest tag. More info at https://hub.docker.com/_/golang/.
No Golang install needed on systems with Docker, both for running or for developing/testing.

Just adding files, setting working dir and running ```make```.

Build:

```
$ docker build -t go-jira .
```

Run:

```
$ docker run --rm -it -e JIRA_API_TOKEN=xxxx -v [your_host_config_yml]:/go/src/gopkg.in/Netflix-Skunkworks/go-jira.v1/.jira.d/config.yml:ro go-jira [other options...]
```
If you're testing locally, perhaps your ```config.yml``` got baked in the previous ```docker build``` and you don't need to ```-v``` mount your host's yml config file.

FYI, there's a still quicker to test, already built image on my Docker Hub. Test it by running:

```
$ docker run --rm -it -e JIRA_API_TOKEN=xxxx -v [your_host_config_yml]:/go/src/gopkg.in/Netflix-Skunkworks/go-jira.v1/.jira.d/config.yml:ro pataquets/go-jira-src [other options...]
```

Using `--rm` causes the container to be deleted after running.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,8 @@
+FROM golang
+
+COPY . /go/src/gopkg.in/Netflix-Skunkworks/go-jira.v1
+WORKDIR /go/src/gopkg.in/Netflix-Skunkworks/go-jira.v1
+
+RUN make
+
+ENTRYPOINT [ "./jira" ]


coryb:
Hey @pataquets, thanks for the PR.  I am not sure about its usefulness though. It does solve some issues for users, but I think it causes other issues.  For example, jira via docker cannot integrate with my os's password keychain or use the `pass` command without a running gpg-agent.  Also for editing issues many people have a specific editor setup that will not be available when run within a container.  

You can get around the password issue by adding the `-e JIRA_API_TOKEN=...` as you mentioned, but that is a not a secure solution, so I have a hard time recommending that approach.  

That said, if we wanted to ship docker images, with alpine and multi-stage docker builds we can minimize the image to be ~40M (or ~12M if we default to vi instead of vim):
```
FROM golang:alpine as builder
RUN apk add --update-cache build-base
WORKDIR /go/src/gopkg.in/Netflix-Skunkworks/go-jira.v1
ADD . .
RUN go build -gcflags="-e" -v -ldflags "-w -s" -o /jira cmd/jira/main.go

FROM alpine
RUN apk add --update-cache ca-certificates vim
COPY --from=builder /jira .
ENTRYPOINT ["/jira"]
```

then this should work:
```
docker build -t jira .
docker run --rm -ti -v $(pwd)/.jira.d:/.jira.d -e JIRA_API_TOKEN=... jira ls
```

pataquets:
Hi @coryb, thanks for your prompt response.

The submitted Dockerfile usage it's not meant for everyone. As you point out, it has several caveats, and most users might prefer using the tool outside Docker. However, for some other (admittedly few) scenarios it might fit. Running for devel/testing/trial cases might be easily done using Docker.

I currently have a Bash alias to run this, feeding the secret from a envrc file I already use for other things. Let me know if you want me to submit an example ```.bash_aliases``` example file.
I can even use the different rcfiles for each dirs/project I already have, similar to ```.jira.d``` dir, but also other logics by just creating different aliases. Gives me a lot of flexibility. However, I must admit that it's a very opinionated way and sure it can be done other ways (I must also admit I'm a Docker freak :wink:) So, secrets on env vars are for me just another secret file to keep, as API tokens would be. I was not even thinking on using pass or agents under Docker (although perhaps could be possible and left as an improvement for those needing/daring).

One of the benefits I'm also looking for is having an always up-to-date running method which installs/updates quickly and easily. I'm very used to it and a few others take advantage of this for some tools, too. My intent for the image is to serve as a minimal image. For instance, if someone wants another editor, it can just create his own AB adding their preferred editor (even with their preferred configs baked in) and link the Automated Build to trigger when the base image is updated (which is what I would do). But without an direct-from-source AB I would have to keep it updated manually. Anyway, your editor comment made me think if at least some editor should be included.

An for the size, I'm +1 with it. If you feel like adding Docker support, I will take another stab at it to make it as smaller as possible using Alpine. I think you've done most of the work already, but I'm OK with testing it and sending a new PR. Also, perhaps the included editor shoud be as small as possible, as you already pointed out.

How does it sounds to you?

pataquets:
@coryb : ping?

robbyoconnor:
Sorry for the dual review.

davidnewhall:
Would you be interested in a full stack build pipeline for this? It's a little crude, but this code will build a multi-arch two-stage Docker image in addition to a homebrew formula, debian and redhat packages. https://github.com/golift/application-builder - I'm using it on a handful of Go projects with great success. I think the best use case is be the homebrew build.  If interested, I'd be happy to put together a pull request, but someone else will likely need to setup Docker Cloud and add deploy key(s) to Travis-CI. I can help with that too.

voiski:
I believe he is using this only to build the binary to compile/test his changes. He wants to avoid the golang installation or even the golang version management. I already went in this path for ruby/java/node/... code XD

It is not for the final user. It is more for development. But, I don't see it required to be in the repo. It is more a personal solution, where you can even embed in you bash_alias as you already mentioned.

```bash
# add it in your aliases
alias go_do="docker run --rm -v $(pwd):/app -w /app golang go"

# Run commands like normal go
go_do build path/to/main.go
go_do test ...
...

# or with make, use your most common user cases
alias go_make="docker run --rm -v $(pwd):/app -w /app golang make"
go_make build lint test
```

nyarly:
@pataquets Are you still interested in this?

pataquets:
Sure, @nyarly.
I somewhat dropped it due to change of tools/priorities, but it regained relevancy lately, so I might pick it up again soon.
Do you have any feedback to share?

I want to stress that my first aim/MVP would be only "package with Docker" (or "add a canonical Dockerfile at a minimum) and artifacts/optimizations/multistages/etc. to come later. I would like to check first if it conflicts with the "use Docker for development workflow" usage. Just an approved Dockerfile that builds, to keep extra maintenance burden low.

PR #215, opened 2018-10-25
add template functions

ThisGuyCodes:
I find myself needing some more basic math when making more dynamic/flexible templates:

- resize columns based on max width of shown values
  - "assignee" has max width of 10, but *this* list only has 4-character names
- dynamically set summary column width based on width of other columns
  - makes templates easier to change
  - works well with dynamic resizing

I'm iffy on adding division, but seemed prudent given the others I added.

### DIFF
--- a/jiracli/templates.go
+++ b/jiracli/templates.go
@@ -108,6 +108,39 @@ func TemplateProcessor() *template.Template {
 		"sub": func(a, b int) int {
 			return a - b
 		},
+		"sum": func(num int, nums ...int) int {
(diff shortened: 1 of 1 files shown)


sstarcher:
Sprig would be a nice addition

coryb:
I ended up adding sprig in #283 so all the template functions you wanted should be available now with the exception of `prod` is added as `mul`.  More docs here:
http://masterminds.github.io/sprig/

I am going to close this PR out now.

PR #240, opened 2019-03-26
API Token auth requires "user" config

jgraglia:
API Token authentication requires the `user ` to be defined in the config


coryb:
Thanks for the PR!  Sorry for the delay.

PR #245, opened 2019-04-21
add pagination to search (closes #211)

### LINKED ISSUE #211: Pagination on Search/List
in late 2017 Atlassian made a change that mandated a limit of 100 results and returned a pagination result in the json.
https://docs.atlassian.com/software/jira/docs/api/REST/7.6.1/#pagination

Currently the start the `start_at` can't be controlled which would allow for a work around.

I was going to try to fix this with a PR but currently I am blocked by https://github.com/Netflix-Skunkworks/go-jira/issues/210

justmiles:


Techcadia:
So I pulled this down and tried it. Works great and as expected. I did find an edge case that I would like the Reviewer to think about.

go-jira defaults `MaxResults` to 500
single API call `MaxResults` to 100

This was never really an issue it just meant that we only got the first 100 results. Now that we loop over this we actually get all the results. The first time I ran it it pulled back 23,000 tickets which took around 70s to return calling the api 230 times. This is probably not what we want at least by default behavior.

This patch is stores the defined MaxResults and then retries items in counts of 100 until reach max.

```
@@ -81,6 +81,8 @@ func Search(ua HttpClient, endpoint string, sp SearchProvider) (*jiradata.Search
        req := sp.ProvideSearchRequest()
        results := &jiradata.SearchResults{}
        issues := jiradata.Issues{}
+       maxResults := req.MaxResults
+
        for {
                encoded, err := json.Marshal(req)
                if err != nil {
@@ -103,10 +105,14 @@ func Search(ua HttpClient, endpoint string, sp SearchProvider) (*jiradata.Search

                issues = append(issues, results.Issues...)
                req.StartAt = len(issues)
+               if maxResults-len(issues) < 100 {
+                       req.MaxResults = maxResults % 100
+               }

-               if len(issues) == results.Total || results.Total == 0 {
+               if results.Total == 0 || len(issues) == results.Total || len(issues) == maxResults {
                        break
                }
+
                resp.Body.Close()
        }
        results.Issues = issues
```

Would love to know your thoughts.

Techcadia:
```bash
###############################################################################
## List 150 closed issues, should be more than 100
###############################################################################

RUNS $jira ls --project BASIC --status Closed --limit 150
IS $(printf $0 | wc -l) -eq 150
```

@coryb I think this would the valid test for this pagination

coryb:
Thanks for the change @justmiles and for the test suggestion @Techcadia, sorry for the delay.

PR #253, opened 2019-05-26
t: rename to _t to fix module support

### LINKED ISSUE #228: Using go-jira as a dependency is broken with go modules ("vgo")
Trying to use `go-jira` in a `go.mod` file like this:

```
module github.com/optile/releasetool

require (
        github.com/Netflix-Skunkworks/go-jira v1.0.20
)
```

`go build` fails with

```
go: extracting gopkg.in/Netflix-Skunkworks/go-jira.v1 v1.0.20
-> unzip /home/jonas/go/pkg/mod/cache/download/gopkg.in/!netflix-!skunkworks/go-jira.v1/@v/v1.0.20.zip: malformed file path "t/.password-store/GoJira/api-token:gojira@corybennett.org.gpg": invalid char ':'
build github.com/optile/releasetool: cannot find module for path gopkg.in/Netflix-Skunkworks/go-jira.v1
```

Do we have a way to get rid of the offending file(s)?

mvdan:
The following two files contain characters which aren't valid in source
files within a Go module:

	t/.password-store/GoJira/api-token:gojira@corybennett.org.gpg
	t/.password-store/GoJira/api-token:mothra@corybennett.org.gpg

The directory only contains test scripts, so it doesn't need to be
included in the module. The simplest way to do that is to start the
directory with an underscore.

Fixes #228.


mvdan:
Darn, the `prove` tool hard-codes the `t/` dir.

mvdan:
Yeah, every time I restart the build different tests fail. I don't know why this is happening. We're discussing it with @coryb over at #go-jira on Slack.

coryb:
Tests failing due to random conditions, the theory is that some other user is running the tests in parallel, we currently do not have locking on the tests so parallel executions will cause conflicts.

vanniktech:
Maybe the master build was racing against the PR build?

PR #255, opened 2019-05-27
README: trim down the content

mvdan:
This README is huge; the document shouldn't take more than two or three
screens. If we need more docs, they should be linked to and live
elsewhere.

For now, start by removing unnecessary sections. Gitter has been
completely unused since 2017, so remove it. The table of contents also
takes a lot of vertical space, when we don't want to have that much
content to begin with.

Tips on forking are also unnecessary, since that's the same for all Go
projects. Moreover, this is out of date now that the project is a Go
module.

Remove the help output, as that can be obtained extremely easily by just
running 'jira -h'.

Finally, remove the v0-v1 changes, as v0 was last released over three
years ago. It's extremely unlikely that anyone still has to upgrade from
the older version.


README.md
```diff
@@ -76,153 +35,6 @@ For example, in bash, adding something along the lines of:
...
-```
-# cat $HOME/.jira.d/config.yml
-#!/bin/sh
-case $JIRA_OPERATION in
-    list)
-      echo "template: table";;
-esac
-```
-
-###### **Custom Commands**
```
vanniktech: Personally, I'd like to keep this section. It was super helpful to me.

mvdan: Not sure I follow - there's still the regular custom commands section. This is the section that covers how the behavior changed between v0 and v1. Like the commit says, v0 is ancient, so I don't think this is relevant anymore.

vanniktech: Oh. Then let's remove it!

PR #263, opened 2019-06-05
Allow reading password from stdin

kojustin:
Allow reading password from stdin in `jira login`. This is useful for
combining the jira tool with other command-line utilities. Fixes https://github.com/go-jira/go-jira/issues/262

### DIFF
--- a/jiracli/password.go
+++ b/jiracli/password.go
@@ -3,6 +3,7 @@ package jiracli
 import (
 	"bytes"
 	"fmt"
+	"io/ioutil"
 	"os"
 	"os/exec"
 	"strings"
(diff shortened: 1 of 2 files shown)


jiracli/password.go
```diff
@@ -95,6 +102,7 @@ func (o *GlobalOptions) GetPass() string {
 }
 
 func (o *GlobalOptions) SetPass(passwd string) error {
+	fmt.Printf("SetPass pw=%s; o.PasswordSource.Value=%s\n", passwd, o.PasswordSource.Value)
```
coryb: probably want to remove this debug output.

coryb: hah, beat me to it :)

kojustin:
@coryb where is another switch on `o.PasswordSource.Value` in `SetPass`. I'm not sure how that's used. This change works for me but i'm not sure if i'm missing something.

coryb:
The SetPass is used for persistent password solutions (keychain or `pass`).  In the case where the users want to use keychain, or pass but has not set it up then `jira` will prompt for the password then call SetPass to have it saved for the next time `jira` is run.  For your use-case it should not be stored, so the SetPass will be a no-op.

PR #267, opened 2019-07-05
update all package references to gopkg.in/jira.v1

coryb:
Removing all the references to the old package location of Netflix-Shunkworks to the new home `github.com/go-jira/jira`, and using `gopkg.in/jira.v1` package.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,4 +1,4 @@
-module gopkg.in/Netflix-Skunkworks/go-jira.v1
+module gopkg.in/jira.v1
 
 go 1.12
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,401 +2,401 @@
 
 ## 1.0.20 - 2018-08-04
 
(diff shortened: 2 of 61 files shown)


go.mod
```diff
@@ -1,4 +1,4 @@
-module gopkg.in/Netflix-Skunkworks/go-jira.v1
+module gopkg.in/jira.v1
```
mvdan: FYI, this is a breaking change; any existing v1 users will have their builds broken if they try to update to a newer v1 version. This is technically breaking semver, as you can only do such a breaking change as part of v2.

Of course, semver is subjective, and you can still merge this; I assume the number of module users are few. Though it would still be nice to do this massive rename as part of a v2, to be on the safe side.

coryb: Okay, makes sense.  I think I will:
* release what is currently on `master`
* branch the current `master` to a `v1` branch,
* update this PR to use `gopkg.in/jira.v2` and tag master with `v2`.

I won't make a v2 release yet, we can decide if there are any other breaking changes that we want to get in before making a v2 release.  This does mean if we want to get a new v1 release out we will have to merge into the v1 branch and then probably rebase onto master.

Does that sound reasonable?

PR #275, opened 2019-09-10
Migrate from gopkg.in to github.com, to support 'go mod' more cleanly

### LINKED ISSUE #228: Using go-jira as a dependency is broken with go modules ("vgo")

mikepea:
I'm not 100% on this PR, so would really like a second pair of eyes from someone who's used to using modules.

This PR removes reference to the old repository and in particular the use of gopkg.in as an import source - it appears to have been preventing go module support from being useful for pinning against a non-released SHA or branch. Essentially, since gopkg.in only pulls vX.Y.Z tagged releases from github, there's nothing there to pull from if you want to use the 'go get {repo}@{git-sha} syntax.

In the process of this, I noticed that #228 was still a problem, despite the move to using '_t' as the prefix for the cli tests. I implemented a temp-file fix to address the problem, so it should now work - at least the tests pass.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,4 +1,4 @@
-module gopkg.in/Netflix-Skunkworks/go-jira.v1
+module github.com/go-jira/jira
 
 go 1.12
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ language: go
 
 go:
   - 1.12.x
+  - 1.13.x
 
 env:
   - GO111MODULE=on
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
-[![Build Status](https://travis-ci.org/Netflix-Skunkworks/go-jira.svg?branch=master)](https://travis-ci.org/Netflix-Skunkworks/go-jira)
-[![GoDoc](https://godoc.org/gopkg.in/Netflix-Skunkworks/go-jira.v1?status.svg)](https://godoc.org/gopkg.in/Netflix-Skunkworks/go-jira.v1)
+[![Build Status](https://travis-ci.org/go-jira/jira/go-jira.svg?branch=master)](https://travis-ci.org/go-jira/jira)
+[![GoDoc](https://godoc.org/github.com/go-jira/jira?status.svg)](https://godoc.org/github.com/go-jira/jira)
 [![License](https://img.shields.io/badge/License-Apache%202.0-blue.svg)](https://opensource.org/licenses/Apache-2.0)
 
 # go-jira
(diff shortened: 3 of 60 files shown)


mikepea:
@coryb are you happy for me to merge this?

go.mod
```diff
@@ -15,6 +15,7 @@ require (
 	github.com/guelfey/go.dbus v0.0.0-20131113121618-f6a3a2366cc3 // indirect
 	github.com/hinshun/vt10x v0.0.0-20180809195222-d55458df857c // indirect
 	github.com/jinzhu/copier v0.0.0-20180308034124-7e38e58719c3
+	github.com/karalabe/xgo v0.0.0-20190301120235-2d6d1848fb02 // indirect
```
coryb: dont think we need this dep, it is only for our cross compile, not sure how to exclude it from the go.mod

coryb:
I have only limited use of go modules, but it seems like to use non-released code you can use the `replace` rule? https://github.com/golang/go/wiki/Modules#when-should-i-use-the-replace-directive
Maybe something like `replace gopkg.in/Netflix-Skunkworks/go-jira.v1 => github.com/go-jira/jira master` syntax in the go.mod?

coryb:
I am not sure of the complications of moving away from gopkg.in, I like the idea of removing the layer of indirection,  but I think it will also break existing users using go modules given comments from this other PR #267.  I think the usage of this is very low, like just admins on this project, so I think we can make this change without to many complications.  Worst-case we can revert if we end up breaking people.

mikepea:
I don't think we're going to affect folk using go modules -- I'm pretty sure modules support was broken until #228 was fixed, and to use modules effectively you'd need to do the replace trick to the current github repo anyway.

I disagree that we need to bump the major version in order to shift away from gopkg.in (re #267) - it's not really a breaking change, it just requires an update to your imports if you wish to use future versions - the code/API remains at v1.

I say let's do it :)

go.mod
mikepea: i've removed this as part of the clean-up rebase.

mikepea:
I'm going to hold off on merging this until I've had a chance to check it with Go 1.13 too -- I've heard it surfaces some module problems that can take a bit of unpicking.

coryb:
You can try to update the .travis.yml from:
```
go:
  - 1.12.x
```
to:
```
go:
  - 1.12.x
  - 1.13.x
```
That should let travis test for you

Issue #281, opened 2019-09-30
When transitioning issues, it only finds the first partial match to the new transition state


pcockwell:
Assume there are 4 states for a ticket - `To Do`, `In Progress`, `PR`, and `Done`

When running `jira transition PR PROJ-1234`, per https://github.com/go-jira/jira/blob/eead13aef14de4f4f4a1393a3f0cb803a03001ff/jiradata/TransitionsFuncs.go#L10-L18 it will only search for the first potential match, which in this case will be `In Progress` because `pr` is contained within `in progress`. 

It would be expected that if another state directly matches the submitted state, that it would prefer the direct match over the earlier partial match

PR #282, opened 2019-09-30
Choose exact transition match if available

pcockwell:
Addresses issue #281 by attempting to choose the exact match if possible.

I am reasonably new to `go`, but believe I have implemented this correctly. That being said, I was unable to determine where the tests for this are written, as well as how to execute those tests. Please feel free to request or make changes as needed, though if you want me to do them you'll have to provide some detail in your instructions so I can understand what you need.

### DIFF
--- a/jiradata/TransitionsFuncs.go
+++ b/jiradata/TransitionsFuncs.go
@@ -11,7 +11,7 @@ func (t Transitions) Find(name string) *Transition {
 	name = strings.ToLower(name)
 	matches := []Transitions{}
 	for _, trans := range t {
-		if strings.Compare(strings.ToLower(trans.Name), name) == 0 {
+		if strings.ToLower(trans.Name) == name {
 			return trans
 		}
 		if strings.Contains(strings.ToLower(trans.Name), name) {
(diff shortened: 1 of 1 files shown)


jiradata/TransitionsFuncs.go
```diff
@@ -9,10 +9,17 @@ import (
 // or nil
 func (t Transitions) Find(name string) *Transition {
 	name = strings.ToLower(name)
+	matches := []Transitions{}
 	for _, trans := range t {
-		if strings.Contains(strings.ToLower(trans.Name), name) {
+		if strings.Compare(strings.ToLower(trans.Name), name) == 0 {
```
coryb: this is more commonly written as `if strings.ToLower(trans.Name) == name {`

pcockwell: Done!

coryb:
not sure what is up with the tests, looks unrelated to your changes, I will look into that when I get a chance.

pcockwell:
@coryb Any idea how long this will be before it makes it into a new release? Just curious so I can check back when ready

Issue #284, opened 2019-10-03
Add gopass support


patrickpichler:
Hi, 
as pass is already supported as password source, i wondered if it would be possible to add gopass support as well.

PR #285, opened 2019-10-03
Add gopass as password source support

patrickpichler:
Add support for gopass as password source

#284

### DIFF
--- a/jiracli/password.go
+++ b/jiracli/password.go
@@ -31,6 +31,13 @@ func (o *GlobalOptions) keyName() string {
 		}
 		return fmt.Sprintf("GoJira/%s", user)
 	}
+
+	if o.PasswordSource.Value == "gopass" {
+		if o.PasswordName.Value != "" {
+			return o.PasswordName.Value
+		}
+		return fmt.Sprintf("GoJira/%s", user)
+	}
 	return user
 }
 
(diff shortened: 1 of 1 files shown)


jiracli/password.go
```diff
@@ -43,6 +50,21 @@ func (o *GlobalOptions) GetPass() string {
...
+				orig := os.Getenv("PASSWORD_STORE_DIR")
+				os.Setenv("PASSWORD_STORE_DIR", o.PasswordDirectory.Value)
+				defer os.Setenv("PASSWORD_STORE_DIR", orig)
+			}
+			if bin, err := exec.LookPath("gopass"); err == nil {
+				buf := bytes.NewBufferString("")
+				cmd := exec.Command(bin, "show", "-o", o.keyName())
+				cmd.Stdout = buf
+				cmd.Stderr = buf
+				if err := cmd.Run(); err == nil {
```
mikepea: Could you add a failure case too please, panic(err) is fine.

jiracli/password.go
```diff
@@ -43,6 +50,21 @@ func (o *GlobalOptions) GetPass() string {
 			if err != nil {
 				panic(err)
 			}
+		} else if o.PasswordSource.Value == "gopass" {
+			if o.PasswordDirectory.Value != "" {
+				orig := os.Getenv("PASSWORD_STORE_DIR")
+				os.Setenv("PASSWORD_STORE_DIR", o.PasswordDirectory.Value)
+				defer os.Setenv("PASSWORD_STORE_DIR", orig)
+			}
+			if bin, err := exec.LookPath("gopass"); err == nil {
```
mikepea: Maybe a warning here for the failure case of gopass not being found? My understanding of the flow is that we'd drop back to the default password behaviour

jiracli/password.go
patrickpichler: I just noticed, that the pass password-source doesn't have error handling at all. Should i add it there too?

mikepea: That'd be lovely, thanks!

patrickpichler: should be done

mikepea:
Thanks for the swift responses!

Could you please do a final run through 'go fmt' so that the indentation is correct, and I'll merge.

mikepea:
LGTM

PR #289, opened 2019-10-21
Add support to switch out password source binary

### LINKED ISSUE #288: Make used gopass/pass binary configurable
It would be nice to make the gopass/pass binary used configurable. This would allow to configure gopass to use for example the QT frontend for pinentry, without changing it for the whole system.

patrickpichler:
Add support to switch out password source binary

    There is now a new configuration entry option `password-source-binary`,
    which allows to use an alternate binary for the gopass/pass password
    source.

    If the option is not specified, we will fallback to `pass` (for `pass`)
    and `gopass` (for `gopass`).

This fixes #288

### DIFF
--- a/README.md
+++ b/README.md
@@ -374,3 +374,19 @@ When `password-source` is set to `stdin`, the `jira login` command will read fro
 ```bash
 $ ./password-generator | jira login --endpoint=https://my.jira.endpoint.com --user=USERNAME
 ```
+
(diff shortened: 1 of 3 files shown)


coryb:
Thanks for the patch, I have rebased for conflicts and renamed the option to `password-source-path` which I feel is more appropriate (minor nitpick).

PR #295, opened 2019-11-15
FIXES #176: templates: default table template was wider than terminal

### LINKED ISSUE #176: Jira table template sometimes gets messed up
I really like the default table template however often times it does not scale well enough and one line is not shown correctly.

Expected:

```
+----------------+------------------------------------------------------------------------------------------------------------------------------------------------+--------------+--------------+--------------+------------+--------------+--------------+
| Issue          | Summary                                                                                                                                        | Type         | Priority     | Status       | Age        | Reporter     | Assignee     |
+----------------+------------------------------------------------------------------------------------------------------------------------------------------------+--------------+--------------+--------------+------------+--------------+--------------+
| foo            |                                                                                                                                                |              |              |              |            |              | 123456789410 |
+----------------+------------------------------------------------------------------------------------------------------------------------------------------------+--------------+--------------+--------------+------------+--------------+--------------+
```

Actual:

```
+----------------+------------------------------------------------------------------------------------------------------------------------------------------------+--------------+--------------+--------------+------------+--------------+--------------+
| Issue          | Summary                                                                                                                                        | Type         | Priority     | Status       | Age        | Reporter     | Assignee     |
+----------------+------------------------------------------------------------------------------------------------------------------------------------------------+--------------+--------------+--------------+------------+--------------+--------------+
| foo            |                                                                                                                                                |              |              |              |            |              | 12345678941011
|
+----------------+------------------------------------------------------------------------------------------------------------------------------------------------+--------------+--------------+--------------+------------+--------------+--------------+
```

This happens on both my Linux as well as on my Mac machine. My hint is that the content of one of the column is too long and there's an off by one error.

dobbymoodge:
This change fixes a bug where the `table` template rendered with a
jagged right edge that extended beyond the rightmost column of the
terminal.

To fix this, the template has been updated to:

* move all the magic numbers for field widths into vars at top of the
  template

* precalculate offsets for the horizontal rule dashing

* increase some field widths


jiracli/templates.go
```diff
@@ -293,14 +293,30 @@ const defaultDebugTemplate = "{{ . | toJson}}\n"
...
-| {{ "Issue" | printf "%-14s" }} | {{ "Summary" | printf (printf "%%-%ds" (sub $w 2)) }} | {{ "Type" | printf "%-12s"}} | {{ "Priority" | printf "%-12s" }} | {{ "Status" | printf "%-12s" }} | {{ "Age" | printf "%-10s" }} | {{ "Reporter" | printf "%-12s" }} | {{ "Assignee" | printf "%-12s" }} |
-+{{ "-" | rep 16 }}+{{ "-" | rep $w }}+{{ "-" | rep 14 }}+{{ "-" | rep 14 }}+{{ "-" | rep 14 }}+{{ "-" | rep 12 }}+{{ "-" | rep 14 }}+{{ "-" | rep 14 }}+
+{{$issuew := int 10 -}}
+{{$typew := int $issuew -}}
+{{$priorityw := int 14 -}}
+{{$statusw := $priorityw -}}
+{{$agew := int 10 -}}
+{{$reporterw := int 25 -}}
+{{$assigneew := int 12 -}}
+{{$dissuew := int (add 2 $issuew) -}}
```
dobbymoodge: `add` seems to return `int64`, but the `sub` template function only wants `int` type args, so there's just a ton of casting going on.

Something like `{{$newval := sub $bar (add 2 $foo) -}}` would give an error like `ERROR Invalid Usage: template: gojira:1:30: executing "gojira" at <$ifoo>: wrong type for value; expected int; got int64`

jiracli/templates.go
```diff
@@ -293,14 +293,30 @@ const defaultDebugTemplate = "{{ . | toJson}}\n"
...
+  | {{ .key | printf (printf "%%-%ds" $issuew)}} | {{ .fields.summary | abbrev (int (sub $summaryw 2)) | printf (printf "%%-%ds" $summaryw) }} | {{.fields.issuetype.name | printf (printf "%%-%ds" $typew) }} | {{if .fields.priority}}{{.fields.priority.name | printf (printf "%%-%ds" $priorityw) }}{{else}}<unassigned>{{end}} | {{.fields.status.name | printf (printf "%%-%ds" $statusw) }} | {{.fields.created | age | printf (printf "%%-%ds" $agew) }} | {{if .fields.reporter}}{{ .fields.reporter.name | printf (printf "%%-%ds" $reporterw)}}{{else}}<unassigned>{{end}} | {{if .fields.assignee }}{{.fields.assignee.name | printf (printf "%%-%ds" $assigneew) }}{{else}}<unassigned>{{end}} |
```
dobbymoodge: I feel like there *must* be a better way to get responsive padding than this pattern: `{{.fields.issuetype.name | printf (printf "%%-%ds" $typew) }}`

I guess we could add a new templating function, something like `"padFit": func(max int, content string) string {...}`. It would look like `abbrev` but also handle the case when `len(content) <= max`, basically moving the `printf (printf ... )` logic into the function using `fmt.Sprintf()` or something...

dobbymoodge: Then - if I understand the template syntax, which I'm still a bit fuzzy on - we would replace the example I quoted with something like `{{.fields.issuetype.name | padFit $typew }}`, which seems like a win

dobbymoodge:
Looks like I have a test to update, lol

dobbymoodge:
OK, I had more than one test to update; didn't realize the `epic-list` template was derived from the `table` template, oops.

If the formatting changes are unwelcome, let me know, they will be easy to tweak if desired.

jiracli/templates.go
coryb: Yeah, this is pretty rough, your solution is probably the best with what we have.   I wonder if we should just add some template functions that correspond to https://github.com/olekukonko/tablewriter to make the table writing easier.

coryb: Hey @dobbymoodge, I played a bit more with `tablewriter`, I think it will work better than this manual craziness that I started with the manual table formatting.  For a proof of concept I was able to write this template:
```
const defaultTableTemplate = `
{{- headers "Issue" "Summary" "Type" "Priority" "Status" "Age" "Reporter" "Assignee" -}}
{{- range .issues -}}
  {{- row -}}
    {{- cell .key -}}
    {{- cell .fields.summary -}}
    {{- cell .fields.issuetype.name -}}
    {{- if .fields.priority -}}
      {{- cell .fields.priority.name -}}
    {{- else -}}
      {{- cell "<none>" -}}
    {{- end -}}
    {{- cell .fields.status.name -}}
    {{-  cell (.fields.created | age) -}}
    {{- if .fields.reporter -}}
      {{- cell .fields.reporter.name -}}
    {{- else -}}
      {{- cell "<unknown>" -}}
    {{- end -}}
    {{- if .fields.assignee -}}
      {{- cell .fields.assignee.name -}}
    {{- else -}}
      {{- cell "<unassigned>" -}}
    {{- end -}}
{{- end -}}
`
```
and I get this now:
```
+------------+---------+------+----------+--------+---------+----------+----------+
|   ISSUE    | SUMMARY | TYPE | PRIORITY | STATUS |   AGE   | REPORTER | ASSIGNEE |
+------------+---------+------+----------+--------+---------+----------+----------+
| BASIC-3679 | summary | Bug  | Medium   | To Do  | 2 hours | gojira   | gojira   |
+------------+---------+------+----------+--------+---------+----------+----------+
```
It auto sizes the columns based on the text, otherwise it will wrap cells to fit within the console width.

So I think I will continue with this approach as the templates seem to be much more readable and will create a PR for it  and ping you.

coryb:
I really appreciate the work you put into this, but we won't be merging it since we just simplified the table creation via #302.

dobbymoodge:
> I really appreciate the work you put into this, but we won't be merging it since we just simplified the table creation via #302.

Oh yeah that's hecka much nicer

PR #296, opened 2019-11-25
Use `stty` when available to set tty width for tests

dobbymoodge:
Checks for `stty` in path and uses it, if available, to set terminal
width (`stty columns 149`). This guarantees that the desired terminal
width is used for tests when the value comes from `ioctl`, since [that
has priority over the `COLUMNS` env var](https://github.com/go-jira/jira/blob/590244947b92c5cfef5dc61b4e1aa931f8da5762/jiracli/templates.go#L101).

This addresses an issue where tests for the output of templates that
use the `termWidth` template function would fail when run in certain
environments - in my case, `bash` running in `rxvt-unicode-256color`
in Fedora 30 - when the terminal width doesn't match `149` like the
test expects.

### DIFF
--- a/_t/env.sh
+++ b/_t/env.sh
@@ -1,4 +1,8 @@
 export COLUMNS=149
+if command -v stty 1>/dev/null 2>/dev/null
+then
+  stty columns 149
+fi
 export JIRA_LOG_FORMAT="%{level:-5s} %{message}"
 export ENDPOINT="https://go-jira.atlassian.net"
 export GNUPGHOME=$(pwd)/.gnupg


_t/env.sh
```diff
@@ -1,4 +1,8 @@
 export COLUMNS=149
+if command -v stty 1>/dev/null 2>/dev/null
+then
+  stty columns 149
```
coryb: This will change my tty params even after the tests exit.  I think we need to preserve/restore the original value to make this sane.  Something like:
```
if command -v stty 1>/dev/null 2>/dev/null
then
    cur=$(stty size  | cut -d " " -f 2)
    stty columns 149
    stty size
    restoreCols() {
        stty columns $cur
        stty size
    }
    trap restoreCols INT TERM EXIT
fi
```

dobbymoodge: That makes sense, though I can't shake the feeling that there's a "right" way to set up a fake TTY for tests like this. I'll do some research when I get a minute tomorrow.

dobbymoodge: although it looks like #302 obviates the need for this PR; I'll double-check tomorrow and close if so.

PR #306, opened 2019-12-22
Adding the option --file to the create jira command

mlerman:
This to allow users to run the jira create command without using the editor. Hence without user interaction. This method of using a file, is more conventional. The input file is strictly a replacement of the editor content. 
I know that it is possible to use the editor with noedit, in combination with the override  options to achieve similar results, and to populate individual fields. But unfortunately, I found that this solution doesn't work in all situations.
example of use:
jira.exe create --file="create_LHR.yml" --project="LHR"


miguelbernadi:
Indenataion seems wonky. May be worth using gofmt on the patches. You'll also have to rebase as there are conflicts now.

nyarly:
Superceded by #422

Issue #314, opened 2020-02-17
Unable to self-assign issues with jira cli


pcockwell:
I've been using the jira cli as part of custom commands I run on my system to simultaneously create a branch based on the ticket name, assign it to myself in Jira, and move the issue to In Progress. Until recently, it has been working quite well. I was on vacation for a few weeks, and upon my return it has stopped working.

- Whenever I run `jira view TICKET_ID`, it returns the ticket details fine
- Whenever I run `jira take TICKET_ID`, I get the following error:
```
ERROR Invalid Usage: 'accountId' must be the only user identifying query parameter in GDPR strict mode.
```

I have been playing around with my API tokens (which are stored in `keyring` on a Macbook), and `.jira.d/config.yml` to see if I can isolate the cause and, along with the error message, have determined the following:
- The new system JIRA is rolling out for GDPR requires using the `accountId` field as a replacement of the `user` field for requests. I have since updated my config to include my `accountId`
- The jira cli will not allow a login using an API token without at least one of the `user` or `login` fields
  - If I remove this field from my config, it incorrectly registers which user I am (thus implying that the `accountId` field is not working for identification purposes)

My current `.jira.d/config.yml` contains the following fields (that have been anonymized):
```
endpoint: https://mycompany.atlassian.net                                                                                                                                      
accountId: MY_PROFILE_ACCOUNT_ID
user: patrick@mycompany.com
password-source: keyring
```

My guess is that the `user` and `login` fields are also the culprit to the GDPR based message, simply because it is complaining that `user identifying query parameter`s are being passed in to the API request.

Is someone able to explain what I am doing wrong, and/or the correct fix for this? I really really like my personal tooling as it increases my productivity, and decreases the repetitive mundane tasks I need to do on a regular basis, and would really appreciate being able to use them again.

Thanks

dvdbng:
Same here, it was working last Friday so it must have been changes in jira API

pcockwell:
@dvdbng I've made an attempt at a fix for this, in the PR above. I've tested the functionality (assign/watch) locally, and it works. Now, we wait on maintainers to see how poor my code is, and what testing related changes need to be made haha. If you are urgently in need of the functionality, you can use those changes, and run `make build && make install` locally. This will give you a `jira` binary located at `$HOME/bin/jira`.

keien:
[...]

PR #315, opened 2020-02-18
Use accountId parameter for GDPR compliant API requests

pcockwell:
Adds `account-id` parameter to config parsing
Uses `accountId` parameter instead of `user` or `name` for `watch`, and `assign` issue commands

I have tested the functionality and can confirm the functionality acts as expected. This being said, I am relatively new to go, and have not gone through the trouble of figuring out how to run the automated tests, or even which tests need to be modified given these changes. Please feel free to advise.

This PR should fix issue https://github.com/go-jira/jira/issues/314

Thanks

### DIFF
--- a/issue.go
+++ b/issue.go
@@ -439,13 +439,13 @@ func RankIssues(ua HttpClient, endpoint string, rrp RankRequestProvider) error {
 }
 
 // https://docs.atlassian.com/jira/REST/cloud/#api/2/issue-addWatcher
-func (j *Jira) IssueAddWatcher(issue, user string) error {
-	return IssueAddWatcher(j.UA, j.Endpoint, issue, user)
+func (j *Jira) IssueAddWatcher(issue, accountId string) error {
+	return IssueAddWatcher(j.UA, j.Endpoint, issue, accountId)
 }
 
-func IssueAddWatcher(ua HttpClient, endpoint string, issue, user string) error {
+func IssueAddWatcher(ua HttpClient, endpoint string, issue, accountId string) error {
 	uri := URLJoin(endpoint, "rest/api/2/issue", issue, "watchers")
-	resp, err := ua.Post(uri, "application/json", strings.NewReader(fmt.Sprintf("%q", user)))
+	resp, err := ua.Post(uri, "application/json", strings.NewReader(fmt.Sprintf("%q", accountId)))
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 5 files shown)


pcockwell:
The newest commit also modifies the output templates to use the `displayName` attribute of the `assignee` or `reporter` instead of the `name` attribute.

Line-Noise:
This is working for me so far. Thanks!

pcockwell:
The newest commit also modifies the output templates to use the `displayName` attribute of the `author` instead of the `name` attribute.

coryb:
Hey, thanks for the PR!  Not sure how to proceed with this, unfortunately, we cannot merge as-is.  This is super frustrating that Atlassian has broken their APIs while not bothering to change the API versions.  One major issue is that this change is only for jira-cloud, but this tool is used a bunch with corporate on-premises Jira installs, so we will need to figure out some sort of compatibility.

We can usually auto-detect jira-cloud since it uses atlassian.net hostnames, but ideally, we would be able to interrogate a Jira API to get information about how the client needs to behave, not sure if something like this exists for Jira.

Also ideally we would be able to figure out what the `account-id` property should be from the login information without having yet another identification property in the config file.  

If anyone has a lead on any of these topics I would appreciate the help.  I think this will require some research to figure out how to solve the compatibility issues, and research takes time which I am very short on these days.

TheKevJames:
> We can usually auto-detect jira-cloud since it uses atlassian.net hostnames, but ideally, we would be able to interrogate a Jira API to get information about how the client needs to behave, not sure if something like this exists for Jira.

I think the "right" way to do this would involve having Jira do things differently in their API. As that idea is probably a non-starter, maybe we could introduce a config option? I imagine we would introduce an `enterprise` argument in the configuration yaml and perhaps fallback to using auto-detection logic if it is unset. For example, the following psuedo-code, which is admittedly a bit of a kludge, could do the trick:
```python
if 'enterprise' not in config:
    # if the user has not given us a hint, do our best to auto-detect via hostname
    config['enterprise'] = bool('atlassian.net' not in config['endpoint'])

use_account_id = not bool(config['enterprise'])
```

> Also ideally we would be able to figure out what the `account-id` property should be from the login information without having yet another identification property in the config file.

This seems possible in principle, since in the worst case we could make a request to their API to find a ticket with `assignee=currentUser()`. I don't know much about the design of this library and whether needing to do this API request would be reasonable design; I also assume its likely that there's a route for this purpose rather than needing to create a hack like that.

> we cannot merge as-is

Totally understand this -- unfortunately, that puts those of us using `jira-cloud` a bit out in the rain. Is there some minimum amount of work we could do to at least unblock those users in the short term? Not sure on this libraries opinion on things like this, but would eg. adding the `enterprise` and `account-id` flags to the config and deprecating them once ~the Jira API gets its act together~ we figure out a stable auto-detection mechanism be reasonable?

coryb:
Closing this issue since we merged #317

PR #317, opened 2020-02-24
update all usage of user.name to user.accountId for privacy migration

coryb:
https://developer.atlassian.com/cloud/jira/platform/deprecation-notice-user-privacy-api-migration-guide/
This PR should replace PR #315 and fix Issue #314

### DIFF
--- a/_t/100basic.t
+++ b/_t/100basic.t
@@ -35,8 +35,8 @@ status: To Do
 summary: summary
 project: BASIC
 issuetype: Bug
-assignee: gojira
-reporter: gojira
+assignee: GoJira
+reporter: GoJira
 priority: Medium
 votes: 0
 description: |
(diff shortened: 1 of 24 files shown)


pcockwell:
I will test this out later today. This looks pretty good for my use cases - thanks for the fix. Am I correct in assuming that having the `user` parameter in my `config.yml`, and also including the `jira-deployment-type` parameter (set to `cloud`) should be sufficient to guarantee the lookup on all calls?

Also, would it be appropriate/relevant to allow the user to (optionally) specify the `accountId` in the `config.yml` so that it would prevent additional API calls where the provided `accountId` is relevant?

coryb:
@pcockwell the user and/or login params options are still required.  The jira-deployment-type option is not necessary, go-jira will call serverInfo api to figure out the deployment type if it is not set.  

Currently, I don't see value in adding an account-id property to the configs, go-jira has to look up users constantly anyway (for assigning to others, adding watchers, setting reporters, etc) so adding an account-id just complicates configuration for little value.

So really there is no change for the user to make with these changes, if go-jira used to work, it should start working again without modifying configs.

pcockwell:
Cool, thanks for the explanation. When I checked out this branch, built, and ran it locally yesterday and today, it's been working just fine for me.

coryb:
Great, thanks for confirming, I will get this merged and released today.

PR #318, opened 2020-02-26
Update README for simpler instructions for Atlassian Cloud users

jrschumacher:


coryb:
Thanks for the updates!

PR #323, opened 2020-03-24
Add support to get all comments for an issue

tjamet:


coryb:
Thanks for the PR!  Not sure why Travis is claiming the build it pending as both builds passed, so I will merge this now.

PR #354, opened 2020-06-18
Extend template functions

kottle:
Added to template extra functions to manipulate int:
My request is to add functions:
add: a+b
div: a/b
mult: a*b
pctVal: (size-val)*100/size


nyarly:
@kottle Any interest in following up here?

Issue #274, opened 2019-09-07
Cut a new release for modules support?


mikepea:
Hey @coryb  - mind if I tag master with the 1.0.21 version, so we've got a real release with modules supported?

coryb:
yeah, sounds good.  I have been meaning to do this for a while.
It has been a while, but I think the release process is:
```
# creates a ./jira binary used later to extract current usage
make

# make all the cross platform binaries, requires docker
make all

# update the CHANGELOG.md
make update-changelog

# update README, create git tag, and push to github
make release
```
Then I copy/paste the changelog snippet into https://github.com/go-jira/jira/releases/new
and attach all the cross platforms binaries in the ./dist directory

Let me know if you run into problems.

mikepea:
The 'make all' step doesn't seem to be working for me at least:

````
Checking for required docker image karalabe/xgo-latest... found.
Cross compiling ....
Building locally ....
Compiling for linux/386...
can't load package: package .: no Go files in /
2019/09/14 22:53:47 Failed to cross compile package: exit status 1.
make: *** [all] Error 1
````

I'm not sure how to get this working

**Comment by coryb** [comment]
Could you try the `make all` again?  I updated the Makefile target to use a fork of xgo more actively supported and theoretically has go module support.

**Comment by mikepea** [comment]
Looks like it works! I've cut the v1.0.21 release - the Makefile process was a little flakey due to NEWVER and CURVER not being calculated correctly (something to do with the v1.0.20 tag?), but I think I've gotten around that by setting them as variables on the command line:

````
CURVER=1.0.20 NEWVER=1.0.21 make update-changelog
CURVER=1.0.20 NEWVER=1.0.21 make release
````

**Comment by coryb** [comment]
Nice, yeah the CURVER and NEWVER can get messed up if you previously ran the release process and it fails somewhere in the middle (after the git tag is applied).  Glad you got it working!

**Comment by ldelossa** [comment]
hey @coryb @mikepea 

I can't do the final step of pushing to master: 
```
❯ git push origin master 
Enumerating objects: 5, done.
Counting objects: 100% (5/5), done.
Delta compression using up to 8 threads
Compressing objects: 100% (3/3), done.
Writing objects: 100% (3/3), 622 bytes | 622.00 KiB/s, done.
Total 3 (delta 2), reused 0 (delta 0), pack-reused 0
remote: Resolving deltas: 100% (2/2), completed with 2 local objects.
[...]

PR #364, opened 2020-08-28
username-deprecation: use email and display names

ldelossa:
this commit deprecates the searching ability by username and
instructs user to provide email or display names in commands.

the username parameter has been deprecated completely from v2 and v3
api

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,14 @@
 
 Simple command line client for Atlassian's Jira service written in Go.
 
+## GDPR USERNAME DISCLAIMER
+
+When this tool was initial written the "username" parameter was widely used in the Atlassian API.
+Due to GDPR restrictions this parameter was been almost completely phased out other then V1 login.
+The "--user" field is still provided as a default global, however moving forward any usage of this field should be phased out in favor of the "--login" option. 
+
+Commands which previously took a username will now expect an email address such as watch, create, assign, etc...
+
 ## Install
 
 ### Download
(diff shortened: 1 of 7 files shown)


ldelossa:
@coryb @mikepea can I get a quick peek? This fixes all tests by deprecating "username" parameter.

ldelossa:
@coryb 

I saw that the tests run a docker container and logins in with "admin". At first I was going to completely remove the "User" value but then I saw that.

This PR doesn't actually change the way login works, the user cli arg is still around, as I thought if I removed it, I'd break the tests which are running against a non cloud deployment I think. 

Either way, the login/logout should work exactly like before, unless you are seeing something I am not.

coryb:
No, I think it is likely fine.  It is hard to say, but there are a few changes like:
```
uri += fmt.Sprintf("?username=%s", user)
```
to:
```
uri += fmt.Sprintf("?accountId=%s", user)
```
that might not work on older Jira servers that dont understand `accountId`.  I don't know, it might be fine, it is hard to track in the api docs when things like accountId were added.  In general the GDPR changes were made in a backwards incompatible way from Atlassian so it hard to know if the code will work in all cases just using the `/api/2` endpoints.  I have no idea why they didnt just deprecate `api/2` and create a `api/3` for these changes.

In any case, since we dont have any testing infrastructure for older Jira servers, I say just go ahead and merge since we know this is causing problems with current Jira Cloud.  If anyone complains hopefully we can get them to help us disambiguate the api requirements for any install they have.

ldelossa:
@coryb ahh totally see your point. 

ok will merge, and lets see if anyone from the community raises issues.

coryb:
@ldelossa when/if you feel like cutting a release there are some instructions in #274

ldelossa:
Thanks Cory, will probably release the fix soon

PR #365, opened 2020-08-29
tests: transition if under review

ldelossa:
this commit attempts to transition any failed test issues if they ticket
is left in "Under Review" state


ldelossa:
@coryb "process.sh" tests were getting stuck due to issue being left in "Under Review" state, most likely from canceled previous tests.

This adds a way to say "okay if we tried to transition from start->stop already, but issues still remain, transition to approve->done"

I know nothing about the workflow in the "Process" project. Are more cautionary transitions necessary in this PR?

ldelossa:
I believe it.

PR #366, opened 2020-08-31
cicd: deflake tests

ldelossa:
this commit removes the parallel test matrix as parallel testing runs
the potention of trampling one another

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,6 @@ addons:
 language: go
 
 go:
-  - 1.12.x
   - 1.13.x
 
 env:


ldelossa:
@coryb 

when both tests are ran in parallel, they trample each other and one may fail depending on the time they start.
I've removed parallel testing. We can just keep bumping the tests to the "oldest supported go version" for now.

ldelossa:
@coryb not a bad idea, any mocks built out yet?

ldelossa:
well, I appear to be wrong actually. the non-parallel test still fails... seemingly intermittently. will do a little bit more exploring.

ldelossa:
unless... same test is running in parallel on "push" and "merge"

coryb:
Yeah, that is certainly possible, not sure if there is a mechanism in travis or maybe github actions that could prevent multiple commits from running tests in parallel.  

btw, I made no progress on the mocks.  Ideally we could run the same tests against mocks and the real integration server, but that means we would have to implement some trivial state machine in mocks to mimic issue transitions etc.  Probably not impossible, but likely a week or two of work.   Hard to justify for this project.

ldelossa:
Fair. Any objection to me attempting a move from travis to gh action?

coryb:
I generally prefer gh actions, travis seems to be losing focus these days.

ldelossa:
Totally agree. Ill take a stab at it this week.

PR #371, opened 2020-09-05
Go Native Testing

ldelossa:
This PR moves the basic.t tests away from passive and into Go's native testing suite. Along the way I ensure all test are parallel safe and will not flake any longer when multiple tests instances are running. 

If this is merged I will continue the migration from passive tests into Go tests. 

This PR also removes travisci in favor of github actions.


ldelossa:
@coryb @mikepea may I get a look at this. You can see the tests passing on my fork:
https://github.com/ldelossa/jira/runs/1082571638?check_suite_focus=true

Issue #299, opened 2019-11-29
Commands without output generate 'ERROR Invalid Usage: JSON Parse Error: EOF'


svenXY:
Hi,

I came across #135 and implemented the suggested custom-command. It works down to the last command where it tries to POST the issue/s to be added to the sprint.

It adds the issue to the sprint successfully, but I keep getting 'ERROR Invalid Usage: JSON Parse Error: EOF'. 

Running this with curl seems to have worked though:
```
curl -D- -u user:password -X POST -d '{"issues":["MYPROJ-1638"]}' -H "Content-Type: application/json" https://jira.company.com/rest/agile/1.0/sprint/2356/issue
HTTP/1.1 204

```

The problem to me seems to be that no output is returned by this command, thus both templates, -t debug and -t request fail as they require some json to be piped into toJson in the template

Looks like a regression to #72 

Is there any easy method to either not spit out the error, or - even better - take the HTTP code and create some output based on the http code

Thanks,
Sven

svenXY:
Update: The problem seems to be to handle the output of the command as the issues are effectively added to the sprint even though errors are shown...

hendrikmaus:
As a quick workaround to get the command to exit with 0, I used:

```
- name: sprint-add
  help: add issue to active sprint for project
  args:
    - name: ISSUE
      required: true
  script: |
    board_id=$({{jira}} req "/rest/agile/1.0/board?projectKeyOrId=$JIRA_PROJECT" --gjq values.0.id)
    sprint_id=$({{jira}} req "/rest/agile/1.0/board/$board_id/sprint?state=active" --gjq values.0.id)
    {{jira}} req -M POST "/rest/agile/1.0/sprint/$sprint_id/issue" "{\"issues\":[\"{{args.ISSUE}}\"]}" > /dev/null 2>&1 || true
```
Where ` > /dev/null 2>&1 || true` is not a pretty thing to do, but makes it work for now.

andrei-a-papou:
I also had to do some ugly stderr redirection to work around this issue...

pylipp:
I had the same issue when deleting an issue link via [this resource](https://docs.atlassian.com/DAC/rest/jira/6.1.html#d2e1926):

```
jira req -M DELETE /rest/api/2/issueLink/47286
```

It shows the program usage, and the message

```
ERROR Invalid Usage: JSON Parse Error: EOF
```

However, the command is executed (i.e. the issue link is deleted).
[...]

PR #372, opened 2020-09-07
Support empty responses in request commands

voiski:
Avoid JSON parser when the response is empty - common cases for HTTP 204 in issues deletion, or moving issues to sprint.

Bellow the example of first running with my branch build and second with the last release.
Screen Shot 2020-09-07 at 12 12 38 PM

### DIFF
--- a/jiracmd/request.go
+++ b/jiracmd/request.go
@@ -72,6 +72,9 @@ func CmdRequest(o *oreo.Client, globals *jiracli.GlobalOptions, opts *RequestOpt
 	if err != nil {
 		return err
 	}
+	if resp.Body == nil {
+		return fmt.Errorf("Empty Response Body")
+	}
 	defer resp.Body.Close()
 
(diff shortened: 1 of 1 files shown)


jiracmd/request.go
```diff
@@ -74,8 +74,17 @@ func CmdRequest(o *oreo.Client, globals *jiracli.GlobalOptions, opts *RequestOpt
 	}
 	defer resp.Body.Close()
```
ldelossa, line 75: This looks good to me, however, there is a chance both body and err can be nil. 

These blocks should actually be: 

If resp.Body == nil  {
  // return error 
}

Then the rest of the code is safe.

ldelossa (changes requested):

ldelossa (approved):
LGTM

voiski:
It fix the https://github.com/go-jira/jira/issues/299

Issue #369, opened 2020-09-04
ERROR reporter: Reporter is required.


wkoszek:
I just fetched the latest revision a1450e879f62be6faaa6cc518521859837ea084f and built it with "make"

I have:

```
endpoint: https://mywrkspace.atlassian.net
user: adam@mycompany.com
password-source: keyring
```

API overall works, e.g.: `jira ls` shows me the tickets fine. I'm attempting to create a ticket in a "backend" project with "BE":

```
jira create --project BE
```

I get dropped to Vim and make a dummy ticket:

```
fields:
  project:
    key: BE
  issuetype:
    name: Bug
  summary: >-
    sample 
  priority: # Values: Highest, High, Medium, Low, Lowest, 
    name: High
  description: |~
    sample
  assignee:
    name: adam@mycompany.com
  reporter:
    name: adam@mycompany.com
```

Error I get:

```
ERROR reporter: Reporter is required.
? Jira reported an error, edit again? Yes
```

Regardless of what I try to put in "reporter" field, I get this error. Should the `jira create` template just work fine? I read README about JIRA username vs login issues and G and also attempted to add:

```
user: adam
login: adam@mycompany.com
```

And I tried following formats in the "reporter":
- adam
- adam@mycompany.com
- "W. Adam Koszek"

Nothing worked. Any hints?

ldelossa:
@wkoszek lets try some low hanging fruit, can you alter your config:

```
endpoint: https://mywrkspace.atlassian.net
login: adam@mycompany.com
user: admin
password-source: keyring
```

login is now the preferred method or accessing the API. This is probably no the issue but worth giving a try. 

Maybe a little tough for me to troubleshoot since when I do "jira create" against my work's server my template includes no "reporter"

coryb:
Probably also worth validating what the jira data schema looks like for `reporter`.  It needs to be a `user` field for the cli to automatically fix the GDPR formatting.  For example this is what is looks like for my projects:
```
$ jira createmeta  --gjq fields.reporter
{
            "autoCompleteUrl": "https://jira.example.com/rest/api/latest/user/search?username=",
            "name": "Reporter",
            "operations": [
                "set"
            ],
            "schema": {
                "system": "reporter",
                "type": "user"
            }
}
```

wkoszek:
```
$ /Users/wk/r_other/jira/jira createmeta --project BE --gjq fields.reporter
{
[...]

PR #377, opened 2020-09-15
update templates with emailAddress and login

keien:
addresses go-jira/jira#369

whether we use `emailAddress` here or `displayName` is probably up for debate, but like I said in the above issue, I find `emailAddress` to be more consistent

I also didn't change `displayName` to `emailAddress` in non-edit commands like `view` - this is probably also up for debate


ldelossa:
I agree, email address is better. You touched on how display names can return multiple values in a query. Email addresses should not if fully qualified.

Issue #290, opened 2019-10-24
bad file descriptor


hkyeakley:
Hello,

I have installed go jira in my home directory on one of my admin systems.

The home directory is NFS mounted from an Isilon storage appliance.

If I try to use go jira from my NFS mounted home directory, I get this.

[my_username@linux_box ~]$ jira ls
usage: jira list [\<flags\>]
alias: jira ls [\<flags\>]

Prints list of issues for given search criteria

Global flags:
--help Show context-sensitive help (also try --help-long and --help-man).
-v, --verbose ... Increase verbosity for debugging
-e, --endpoint=ENDPOINT Base URI to use for Jira
-k, --insecure Disable TLS certificate verification
-Q, --quiet Suppress output to console
--unixproxy=UNIXPROXY Path for a unix-socket proxy
--socksproxy=SOCKSPROXY Address for a socks proxy
-u, --user=USER user name used within the Jira service
--login=LOGIN login name that corresponds to the user used for authentication

Optional flags:
-t, --template=TEMPLATE Template to use for output
--gjq=GJQ GJSON Query to filter output, see https://goo.gl/iaYwJ5
-a, --assignee=ASSIGNEE User assigned the issue
-c, --component=COMPONENT Component to search for
-i, --issuetype=ISSUETYPE Issue type to search for
-l, --limit=LIMIT Maximum number of results to return in search
-p, --project=PROJECT Project to search for
-n, --named-query=NAMED-QUERY The name of a query in the queries configuration
-q, --query=QUERY Jira Query Language (JQL) expression for the search
-f, --queryfields=QUERYFIELDS Fields that are used in "list" template
-r, --reporter=REPORTER Reporter to search for
-S, --status=STATUS Filter on issue status
-s, --sort=SORT Sort order to return
-w, --watcher=WATCHER Watcher to search for

ERROR Invalid Usage: Failed to save cookie file: bad file descriptor

I have verified that the /home directory is mounted RW and that I have RW to my home directory.

I tried setting my home directory to /var/tmp which is running on LVM via a local disk.

When I run go jira from here, it works fine.

[my_username@linux_box /var/tmp/my_username~]$ jira ls
XYZ-001:	Issue #1
XYZ-002:	Issue #2
XYZ-003:	Issue #3
XYZ-004:	Issue #4
XYZ-005:	Issue #5
XYZ-006:	Issue #6
XYZ-007:	Issue #7
XYZ-008:	Issue #8
XYZ-009:	Issue #9
[...]

PR #387, opened 2020-10-12
Move the session cookie to $XDG_RUNTIME_DIR

knedlsepp:
Before this commit the session cookie and its corresponding lock file
would be placed in the user's home directory. Sadly this breaks all
setups that configure the user's home to live on a NFS mount. File
locking on NFS mounts is something that doesn't work reliably and thus
causes some users to experience issues like this:

    Failed to save cookie file: bad file descriptor

To counteract, this commit instead stores this data in the
$XDG_RUNTIME_DIR which is a directory that is usually located on the
local disk or even located in runtime memory.

This commit will most likely slightly change the behaviour of how
persistent the session data will be stored since often $XDG_RUNTIME_DIR
will be cleared on logout of the user.

Fixes: #290

### DIFF
--- a/cmd/jira/main.go
+++ b/cmd/jira/main.go
@@ -45,7 +45,12 @@ func main() {
 		panic(jiracli.Exit{Code: 1})
 	}
 
-	o := oreo.New().WithCookieFile(filepath.Join(jiracli.Homedir(), configDir, "cookies.js")).WithLogger(&oreoLogger{log})
+	if err := os.MkdirAll(filepath.Join(jiracli.Cookiedir(), configDir), 0700); err != nil {
+		log.Errorf("%s", err)
+		panic(jiracli.Exit{Code: 1})
+	}
+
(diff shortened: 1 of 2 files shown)


hrivera-ntap:
Any chance this could get reviewed and merged?

knedlsepp:
Still need this patch locally. Would appreciate if somebody takes a look. :-)

knedlsepp:
Thanks for the review. In fact I did not consider macOS or Windows when writing this. This PR should keep the existing behavior of writing to the user's home directory in case that environment variable is unset. Another alternative would be to rely on some other package to provide this functionality. How about depending on https://pkg.go.dev/github.com/zchee/go-xdgbasedir#section-readme for this?

knedlsepp:
> @knedlsepp Sorry to be returning to this after so long - are you still interested in this issue?

I'm still locally applying this patch for my use case, so one could say I'm still interested. I don't have much experience in go, so probably won't find time to give the alternative I mentioned in https://github.com/go-jira/jira/pull/387#issuecomment-1312526260 a shot.

PR #398, opened 2020-11-30
Load config on "jira browse"

nenad:
This is done in order to ensure consistency with the other commands on how the configuration files are loaded. I found out about it when I had set up a global project, but the "jira browse 1234" command complained about missing project.


plundra:
This works as intended for me and would be appreciative if was merged :-)
Tested for exact use-case mentioned, so it works like `jira view -b`

PR #406, opened 2021-01-18
feat: add fromFile in create

denysvitali:
This PR adds the `--fromFile` argument to `jira create`. This is a pretty useful addition when it comes to generating YAML files from outside of jira-cli.

My use case is being able to use a YAML file generated from another tool / script, so that jira-cli can create the issues using that file as an input.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -23,7 +23,7 @@ require (
 	github.com/imdario/mergo v0.3.7 // indirect
 	github.com/jinzhu/copier v0.0.0-20180308034124-7e38e58719c3
 	github.com/kballard/go-shellquote v0.0.0-20180428030007-95032a82bc51
-	github.com/kr/pretty v0.1.0
+	github.com/kr/pretty v0.1.0 // indirect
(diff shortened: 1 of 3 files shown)


denysvitali:
@renelehmann This messes up this Pull Request...
You probably shouldn't run your compliance script against forks

renelehmann:
> @renelehmann This messes up this Pull Request... You probably shouldn't run your compliance script against forks

Sorry @denysvitali  I wasn't aware of a PR based on the fork master branch. I just reverted adding these files now.
These files would be needed for FOSS compliance checks and if these files are missed, the public and forked repository will be set as "archived" after an "internal" compliance check soon.

denysvitali:
I don't have permission over that repo anymore, can you maybe create a feature branch instead, so that I can switch branches in this PR?

renelehmann:
@denysvitali I created a new branch based on your commit. Can you please try to switch your PR to this branch.
https://github.com/swisscom/jira/tree/denysvitali-feature-createfromfile

PR #409, opened 2021-02-04
Add command to show versions for a project

kamilwaz:
It adds `jira versions` that list versions for a project. The command takes the same arguments as `jira components`.


pkulev:
@coryb @ldelossa @mikepea bump please

georgettica (approved):
Looks ok, should test and see it works as expected.

Also maybe adding a test for this would help speed it up

kamilwaz:
I tested it locally and it worked as expected. I can add a test but I'm not sure where - should it be added to `test` or `_t` (or both)? Is there any existing test that I can use as template?

pkulev:
👍 🤣

PR #410, opened 2021-02-17
Add releases command to list project versions

froque:


charles-dexter-ward:
It will be very useful when merged. Now I have been forced to implement it in my code.

PR #411, opened 2021-02-18
Remotelink (web link) support

jmtd:
This adds a new command "remotelink" which permits adding a remote (or
web link) to an issue.

This is (probably) pretty rough. This is my first attempt at writing Go! I have not
yet added verbs for fetching existing remote links, removing remote links, etc.

It's perfectly possible to add two or more remote links to an issue that are identical,
and it might be nice to detect and prevent that.

### DIFF
--- a/issue.go
+++ b/issue.go
@@ -338,6 +338,40 @@ func LinkIssues(ua HttpClient, endpoint string, lip LinkIssueProvider) error {
 	return responseError(resp)
 }
 
+// https://developer.atlassian.com/cloud/jira/platform/rest/v2/api-group-issue-remote-links/#api-rest-api-2-issue-issueidorkey-remotelink-post
+func (j *Jira) LinkRemoteIssue(issue string, url string, title string) error {
+	return LinkRemoteIssue(j.UA, j.Endpoint, issue, url, title)
+}
+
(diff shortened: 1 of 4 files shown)


issue.go
```diff
@@ -338,6 +338,40 @@ func LinkIssues(ua HttpClient, endpoint string, lip LinkIssueProvider) error {
...
+
+func LinkRemoteIssue(ua HttpClient, endpoint string, issue string, url string, title string) error {
+        req := struct {
+            Object struct {
+                Url   string `json:"url"`
+                Title string `json:"title"`
+            } `json:"object"`
+        }{}
+        req.Object.Url = url
+        req.Object.Title = title
```
taliastocks, line 354: If you ever pick this up again, would you mind adding support for other fields? e.g. https://github.com/phacility/phabricator/blob/5720a38cfe95b00ca4be5016dd0d2f3195f4fa04/src/applications/doorkeeper/worker/DoorkeeperJIRAFeedWorker.php#L230-L249

jmtd: I'm unlikely to return to this, to be honest; I think the odds of it being merged are close to zero. Although I'm still using a build with this in it personally. on my TODO list is to look at https://github.com/ankitpokhrel/jira-cli and see if that's a better candidate going forward.

PR #424, opened 2021-05-05
Update prepare-release.yml

georgettica:
this should bump versions and make the code be more compliant

the issue we had before in prev PR seems to be because of environments we are not setting

### DIFF
--- a/.github/workflows/prepare-release.yml
+++ b/.github/workflows/prepare-release.yml
@@ -25,12 +25,13 @@ jobs:
       - name: Changelog
         shell: bash
         run: |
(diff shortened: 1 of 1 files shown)


georgettica:
might solve #423

georgettica:
@mikepea are you still an active maintainer in this repo?
if so I would love to get the +1 from you on this

tested partially on my machine and it looked to do what I wanted (just saw the binary is in the same name and location.. didn't check api changes yet

ldelossa:
This looks good to me.

ldelossa (approved):

PR #425, opened 2021-05-12
fix(actions): fix release.yml

georgettica:
this is a WIP and should solve the promotion problem


georgettica:
@ldelossa it's as close as I can get to testing it works.. care to merge and see if it fixes the deploy problem?

PR #438, opened 2021-08-03
Add authentication-method: bearer-token

rvl:
Hi, thanks for the neat tool.

For some reason, our private Jira installation requires the Bearer authentication scheme for API tokens, instead of Basic authentication. I don't know the reason - perhaps it's because there is SAML configured.

Anyway, this patch adds support for such configurations.

Cheers


catskul:
@georgettica this look good for merge?

georgettica:
I guess so, but I am not sure how I would test this (as my org doesn't use the API bearer token)

I'll try to test this and come back to you

catskul:
I'm testing this right now, and found one issue. Looks like `password.go:136` needs an update to include `bearer-token` Not sure what the right thing to do here is, I suppose I can make a pull request to @rvl 's branch

catskul:
Found a few others. I'm going to submit a fix branch in a few min.

catskul:
Took way longer than expected, but: https://github.com/go-jira/jira/pull/447

catskul:
Thanks @mikepea !

PR #442, opened 2021-08-24
add --passwordsource global option

chuckliu1979:
for automation testing and/or CI/CD, it's good to have pipe login from stdin. at present we need add "password-source: stdin" in .jira.d/config.yml, it's not convenient. this pull request would better :

echo "password" | jira login --passwordsource stdin


go.mod
```diff
@@ -48,3 +48,5 @@ require (
 	gopkg.in/op/go-logging.v1 v1.0.0-20160211212156-b2cb9fa56473
 	gopkg.in/yaml.v2 v2.2.2 // indirect
 )
+
+replace github.com/go-jira/jira => github.com/chuckliu1979/jira v1.0.30 //indirect
```
catskul, line 52: oops, this probably shouldn't stay in.

mikepea: +1 we can't merge with this in

catskul (changes requested):
Need to fix the `replace` line in go.mod at least. Did not review the rest.

PR #447, opened 2021-11-21
Fix rvl auth bearer token branch

catskul:
Some fixes to https://github.com/go-jira/jira/pull/438


georgettica:
Wow, that is deeper than the original

catskul:
> Wow, that is deeper than the original

Yeah, it turns out most of the places we were comparing to "api-token" we meant token.

One example was that despite having JIRA_API_TOKEN set it was still asking me for a password.

catskul:
One thing that could be taken a step further is that the password store could be made to differentiate between `bearer-token` and 'auth-token' but that felt like a step to far for this PR

catskul:
@coryb, @ldelossa, @mikepea could we get eyes on this PR?

mikepea (approved):
I don't have a local JIRA instance to test this against, but the code appears solid and it matches to the documentation at https://confluence.atlassian.com/enterprise/using-personal-access-tokens-1026032365.html#UsingPersonalAccessTokens-UsingPATs

Merging!

PR #454, opened 2021-12-08
fix(watch): make delete watcher work

georgettica:
the apispec noted it should use 'username' and not 'accountId'

### DIFF
--- a/issue.go
+++ b/issue.go
@@ -499,7 +499,7 @@ func (j *Jira) IssueRemoveWatcher(issue, user string) error {
 
 func IssueRemoveWatcher(ua HttpClient, endpoint string, issue, user string) error {
 	uri := URLJoin(endpoint, "rest/api/2/issue", issue, "watchers")
-	uri += fmt.Sprintf("?accountId=%s", user)
+	uri += fmt.Sprintf("?username=%s", user)
 	resp, err := ua.Delete(uri)
 	if err != nil {
 		return err


georgettica:
@catskul can you look?
at least peer review can help 🤞

nyarly:
@georgettica Can you link the spec? It looks straightforward but it's hard to comment without context

issue.go
```diff
@@ -499,7 +499,7 @@ func (j *Jira) IssueRemoveWatcher(issue, user string) error {
 
 func IssueRemoveWatcher(ua HttpClient, endpoint string, issue, user string) error {
 	uri := URLJoin(endpoint, "rest/api/2/issue", issue, "watchers")
-	uri += fmt.Sprintf("?accountId=%s", user)
+	uri += fmt.Sprintf("?username=%s", user)
```
nyarly, line 502: Per the docs:
```
username
string
This parameter is no longer available. See the deprecation notice for details.
```
Jira is transitioning wholesale to the (sadly opaque) `accountId`.

nyarly (changes requested):
I think in general we should be switching to `accountId` for anything but privately hosted instances. What bug is this addressing?

issue.go
nyarly: Or something. When I was first starting out with go-jira this was a new issue, but the way the Jira ecosystem is, we probably need a better solution. Could we a) set up a maintainer team or b) enable discussions here - I think this is a bigger design issue that we can usefully resolve in PR comments.

Issue #459, opened 2022-04-07
Would you accept a PR to support the password-source: keyring config on Windows?


bkane-msft:
When `password-source: keyring` is set in my config and I run on Windows, I get the following error:

```
PS C:\Users\bkane> jira login
Keyring is not supported for Windows, see: https://github.com/tmc/keyring
```

I think this keyring support in Windows can easily be added by replacing https://github.com/tmc/keyring with https://github.com/zalando/go-keyring in the `keyringGet`/`keyringSet` code.

Would you accept a PR with this code change and updated docs? `jira` would be easier to use on Windows and I could share the same config between Windows and Mac :)

chipbite:
It seems that PR was not accepted and merged, right? 
Wanted to try this on windows, but one still cant use windows keyring, I guess?

bkane-msft:
That PR was not merged, and, two years later, I doubt that it will be. You could compile locally from the branch and run that I suppose.

PR #460, opened 2022-04-20
Replace github.com/tmc/keyring  with github.com/dougEfresh/keyring

dougEfresh:
github.com/tmc/keyring uses an archived and outdated dbus lib: https://github.com/guelfey/go.dbus

I have forked keyring and updated to use github.com/godbus/dbus/v5

https://github.com/godbus/dbus


dougEfresh:
I do have an open PR to tmc:
https://github.com/tmc/keyring/pull/21 if you want to wait and use that instead

georgettica:
This seems like a trivial change, let's wait and see the response..
If the upstream changes we will bump up.
If not we will add the fork and track the PR upstream

georgettica (changes requested):
I am not sure this error is problematic, but can you take a look to see if it reproduces locally / there is a PR to fix it?
If so I will merge it before merging this

georgettica:
I think a local https://github.com/nektos/act should make sure it works.
Ty! (Didn't check it works really)

bkane-msft:
Unfortunately, https://github.com/tmc/keyring does not support Windows. In addition, the last commit to tmc/keyring was 2017, so it remains to be seen whether any one is around to merge https://github.com/tmc/keyring/pull/21 .  

I made #459 to propose using https://github.com/zalando/go-keyring for keyring management in `go-jira` because it seems to be maintained. it looks like [they updated dbus support](https://github.com/zalando/go-keyring/commits/master) on  Feb 23, 2022 ). Importantly for me, zalando/go-keyring  already supports Windows.

@dougEfresh , what are your thoughts on using zalando/go-keyring for go-jira? Then you get updated dbus support "for free" without needing to maintain a fork and I get Windows support "for free" as well.

As I stated in #459  I'm happy to do the work integrating zalando/go-keyring in go-jira. I *think* it'll be easy. I'd also be super happy if you want to do this integration work @dougEfresh :D

bkane-msft:
Just made https://github.com/go-jira/jira/pull/467 to use `zalando/go-keyring` for keyring support (updates Linux support, adds Windows support).

PR #466, opened 2022-08-16
Added the more info if command get failed

atul-bhardwaj-sp:
sometimes it gives html rather then the json and due to this it's very hard to get what is the output only i get
Invalid Usage: invalid character '<' looking for beginning of value, may be some problem with jira api call and when i rerun the command it works fine  but it's good to have output when it gives any error and also if i use -vvv more verbose it's show authorization bearer token on that and i don't want  to show it.

### DIFF
--- a/issue.go
+++ b/issue.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"log"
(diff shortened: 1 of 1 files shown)


rishabh-saxena-sp (approved):
+1

Issue #481, opened 2022-12-17
404 using jira create to on-premise Jira Server 9.x


prpht9:
What happened?
https://confluence.atlassian.com/jiracore/createmeta-rest-endpoint-to-be-removed-975040986.html

Jira 9.0 moved the parameters on api/2/issue/createmeta into path values
Need to allow both versions to function
Need a configuration item to indicate which url/method to call
Need to write the new method to conform to the new api method format
Need new jiradata type because the api endpoint changed the output json schema

slipstream to create jiradata types
slipscheme -stdout schema/IssueTypes.json > jiradata/IssueTypes.go

Old API Doc: https://docs.atlassian.com/software/jira/docs/api/REST/7.2.7/#api/2/issue-getCreateIssueMeta

New API Doc: https://docs.atlassian.com/software/jira/docs/api/REST/9.0.0/#issue-getCreateIssueMetaProjectIssueTypes

What did you expect to happen?
jira create should work against Jira Server 9.x

How can we reproduce it (as minimally and precisely as possible)?
setup .jira.d/config.yaml to point to a Jira Server 9.x

jira login
jira create
This should produce a 404

Minimal code helps us to identify the problem faster.

Anything else we need to know?
I have forked go-jira and attempted to create a fix for this. It is partially there. I at least figured out how to get it to work. But I don't know how you all would prefer to setup the configuration to choose between the old way and the new way. Also, I'm not that much of an expert with golang. So I may not have changed the code correctly. I'm hoping for some peer review and guidance so we can get this working for everyone.

Here is the PR I created in my fork: https://github.com/prpht9/go-jira/pull/1
I created the TODO.md file to track notes about my work
I created the schema/IssueTypes.json in order to use slipstream to create the jiradata/blah.go file(s)
These can be removed before merging but I felt they communicate good stuff
I can create a PR from my fork to this upstream repo if/whenever you wish. Just let me know.

Your Environment
On-Premise Jira Server 9.2, forked go-jira project, changed code, rm .jira.d/config.yml from project because it was locked to dev params, ./jira create

Include as many relevant details about the environment you experienced the problem in

go-jira version (git tag or sha): latest master
Go version (go version): 1.19.4
Jira type (cloud or on-premise): on-premise
Jira version / Api version: Jira 9.2 / api v2

philipsd6:
@coryb @mikepea @ldelossa @georgettica can we get this fixed? The more time that goes by the less this project will be useful, as Jira instances are kept up to date...

[...]

Issue #490, opened 2023-03-13
createmeta issue on jira 9.x - "Issue Does Not Exist"


k-wall:
After a recent JIRA upgrade, go-jira `jira create` nows fails for me.  Turning up logging shows:

```
DEBUG Request 1: GET /rest/api/2/issue/createmeta?projectKeys=MGDSTRM&expand=projects.issuetypes.fields HTTP/1.1

DEBUG Response 1: HTTP/2.0 404 Not Found

{"errorMessages":["Issue Does Not Exist"],"errors":{}}
```

k-wall:
I found https://github.com/go-jira/jira/pull/480 which lead me to this user's work: https://github.com/prpht9/go-jira/commit/3c017959dcfb064f3b2e915d4c7b618853b23a61

k-wall:
Dup of #481

PR #502, opened 2024-04-02
fix createmeta issue on jira 9.x

prpht9:
* Jira changed the format of the call to createmeta

 * config.yml should default to old way
 * add a jira server version variable to config.yml to tell cli which method call to make
 
 * fixes issue #481 , confirmed by creator of #490


jmtd:
I can't formally review this, but it has fixed the problem for me. From a merging POV, the TODO.md is superfluous, and the additional config value should be documented. I didn't need to set it in order to file an issue, and I'm still unsure what circumstances I'd need to set it and to what.

darthwalsh:
@prpht9  I was trying out this PR, and hit a small issue with overriding `components` using `jira create`.

My jira server (v9.12.12) responded to `/rest/api/2/issue/createmeta/ABC/issuetypes` without any `.allowedValues` in the response. I needed to query a specific issue type e.g. `/rest/api/2/issue/createmeta/ABC/issuetypes/3` and then `.allowedValues` showed up for all fields.

This led to the default [`create` template](https://github.com/go-jira/jira/blob/4263bd24f9e9c702a92358c5cd7ce0ddd711df4c/jiracli/templates.go#L486) not having any `.meta.fields.components.allowedValues` , which leads to CLI `jira create --override components=Something` getting ignored. It's might be fair to call "components gets ignored" a separate preexisting issue, but it's exposed by this PR's fix.

As a workaround, I created a custom creation template that hardcodes
```yaml
  components:
    - name: {{ .overrides.component }}
```

prpht9:
@darthwalsh  Apologies for the delayed response.  I think I had to do something similar but didn't know where to fix this in the repo.  I'm not an active golang dev.  If you know where to fix this in the code, go ahead and kick up a PR.  I'll review it and merge if it pans out.

PR #510, opened 2025-09-02
Fix jira api

j-mcnally:
The api originally used is deprecated, this updates it.


DBX12 (approved):
I'm not a maintainer of the project but for what it's worth, I consider the changes in good faith and expect them to solve the issue.

jaymzh:
This breaks on-prem users who aren't on the absolute-latest. :(

PR #514, opened 2026-06-18
Add an api-version flag

jaymzh:
While Atlassian cloud does not support v2, not everyone is on
Atlassian cloud. :)

Fixes: #510

Signed-off-by: Phil Dibowitz <phil@ipom.com>


jaymzh:
@coryb - mind having a look at this?

